PR #24, opened 2020-01-10
Update app.ts

hanicker:
I had some problems due to Request Too large, fixed with this change as suggested in https://github.com/expressjs/body-parser/issues/235


agoldis:
@hanicker I forgot to thank you for the contribution! Thank you :) 👍🏻

PR #61, opened 2020-04-27
Fix runs query resolver to match schema

### LINKED ISSUE #62: Fix runs query resolver
The query runs resolver return value does not match the defined schema.

I have implemented a fix in #61

xtroncode:
The query `runs` returns an error stating `Expected Iterable, but did not find one for field Query.runs.`  as the object returned from the resolver does not match the Query schema. So I fixed the getAllRuns api to use the `fullRunReducer` instead of the `runFeedReducer`
fixes #62


agoldis:
Will look into that later today

PR #63, opened 2020-04-28
implement delete run mutations and delete button on dashboard

xtroncode:
Implement graphql mutations for
- deleteRun
- deleteRuns
- deleteRunsInDateRange

Implement delete button on dashboard
- invoke deleteRun mutation on delete button click
- clear queries from cache on delete success

image


xtroncode:
@agoldis  
I have finished implementing the mutations and also added a delete button on the UI.
I was not sure where to place the `deleteRun.graphql` file and hence I have created it along side other `.graphql` files.
Please review.

agoldis:
Will review later today

agoldis:
Nice job @xtroncode!

I have added confirmation dialog for the deletion on top of your changes, in addition to some configuration updates.

While testing it I have discovered that better approach would be just marking runs as deleted, without actually deleting them.

That would prevent the whole bunch of issues - e.g. deleting an ongoing run causes exceptions on director.

I won't request the change now, but if you feel like refactoring the implementation, please do.

xtroncode:
@agoldis Thanks for the review. I had thought about having a confirmation dialog, thanks for implementing it. 
I agree with you that it will cause an issue, when deleting runs that are in progess. 
Is there a flag to determine whether a run is in progress or completed ? We can hide the button based on that.

If not, then the approach you have suggested works fine, we can then clean data periodically.

Thanks.

agoldis:
There's no explicit running / not running state for a run - there's a list of specs, some of the might be claimed and some not. 
However, a run with non-claimed spec might just a stuck run i.e. it is not necessarily still running.

The original cypress dashboard implements "timeout" for each run - if not specs were claimed within specific time window since the last claim, the run is considered "done".

We can apply similar logic, but I am not sure it worth the effort.

PR #65, opened 2020-04-29
Add cypress wrapper

janineahn:
Regarding #8, this adds a helper script for rerouting cypress via your hosts file.

### DIFF
--- a/bin/sorry-cypress.js
+++ b/bin/sorry-cypress.js
@@ -10,7 +10,7 @@ const wording = {
     envNotSet: 'Please check your environment variable: '
 }
 
-const cypressArray = [
+const cypressCommand = [
     'run',
     '--record',
     '--parallel',
@@ -20,7 +20,8 @@ const cypressArray = [
 
 // exit on windows
 if (process.platform === 'win32') {
-    process.exit(wording.noWindows);
+    console.error(wording.noWindows);
+    process.exit(1);
 }
 
 // check env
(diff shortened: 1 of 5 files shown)


agoldis:
Thanks a lot @janineahn  and @redaxmedia! 

---

I am planning to have this tool in a separate repo. I will approve and merge the PR to persist your contribution, and then will extract it to a separate repo and release to npm. Are you ok with that?

redaxmedia:
Haven the CLI inside another project totally makes sense cause using `npx sorry-cypress` would fetch to much unneeded files and dependencies.

Our solution comes with some drawbacks that should be considered. It worked for us but needs rework to work for everyone properly!?

1. Only works with OS that has `etc/hosts`
1. Only works with HTTPS on the target machine*
1. Only works with port 443 on target machine
1. Has fixed arguments for cypress run

*Could properly fixed via HTTPS inside Docker but that's seriously a big task

bin/sorry-cypress.js
```diff
@@ -0,0 +1,81 @@
...
+const cypressArray = [
+    'run',
+    '--record',
+    '--parallel',
+    '--key=' + process.env.SORRY_CYPRESS_RECORD_KEY,
+    '--ci-build-id=' + process.env.SORRY_CYPRESS_BUILD_ID
+];
+
+// exit on windows
+if (process.platform === 'win32') {
```
redaxmedia, line 22: Exit on Windows can be removed as hostile supports it too

redaxmedia:
@agoldis You are welcome

Issue #110, opened 2020-07-22
Turn off autorefresh by default


tallkid24:
The autorefresh is nice but every second its retrieving MBs of data which if you accidentally keep the dashboard up, will eat up your datacap. Would be nice if it was turned off by default and then you could switch it on.

![image](https://user-images.githubusercontent.com/52414395/88202681-d9912d00-cc0e-11ea-9534-c1a8df8b9615.png)

agoldis:
😃  @jhicken FYI

jhicken:
@agoldis should we add an env var for that? Or we could do a localstorrage setting in the UI. 

Thoughs?

agoldis:
- env is too heavy - ppl will need to restart / redeploy the service
- local storage is a good option, but people will still need a UI control for that.

May be a toggle in the dashboard with `off` by default? 

@tallkid24 @jhicken  how does that sound to you?

KyleThenTR:
Sounds good to me

jhicken:
Yup sounds good.

agoldis:
@jhicken I have temporary disabled polling for root view due #111 

I will work later polling-friendly pagination and toggle as discussed on #110

agoldis:
Resolved by #122

PR #112, opened 2020-07-23
Disable polling to RunsView

### LINKED ISSUE #111: Old runs loaded with Load More button disappear after 1 second
**Repro steps:**

1. Open dashboard with more than 5 runs recorded
2. Click Load More button to see old runs
3. Wait a while, check if old runs are visible

**Current result:**
Old runs disappear after 1 second, probably due to the autorefresh feature

agoldis:
Resolves #111

### DIFF
--- a/packages/dashboard/src/views/RunsView.tsx
+++ b/packages/dashboard/src/views/RunsView.tsx
@@ -1,8 +1,8 @@
+import { useApolloClient } from '@apollo/react-hooks';
+import { Button } from 'bold-ui';
 import React from 'react';
 import { RunSummary } from '../components/run/summary';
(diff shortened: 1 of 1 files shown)


agoldis:
@jhicken I have temporary disabled polling for root view due to that issues. I will work later polling-friendly pagination and toggle as discussed on #110

jhicken:
Cool thanks for the heads up. I was planning on putting a toggle on the page to enable/disable polling. I will probably have time to do that on tuesday. If you have not gotten to it by then. Pretty slammed today at work and just left for a vacation.

agoldis:
@jhicken np, we are all pretty busy :)

PR #122, opened 2020-08-07
Adding auto reload toggle and tooltips on breadcrumbs

jhicken:
I did not do anything with the auto refresh on the runs page. Due to the stuff with load more. If we switch that to pagination hopefully that will be simple to implement.

Screen Shot 2020-08-07 at 5 18 42 PM
Screen Shot 2020-08-07 at 5 17 55 PM

Also on a side note for global state are we planning to just use apollo cache? If so maybe I can migrate the local storage updating and such to somewhere in the application bootstrap.

### DIFF
--- a/packages/dashboard/src/components/layout/header.tsx
+++ b/packages/dashboard/src/components/layout/header.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import truncate from 'lodash.truncate';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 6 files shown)


jhicken:
This should address #110

agoldis:
> Also on a side note for global state are we planning to just use apollo cache? If so maybe I can migrate the local storage updating and such to somewhere in the application bootstrap.

I don't really care, local storage seems good to me. If you have any particular interest in gql cache, up to you :)

PR #127, opened 2020-09-01
Add heuristics to run view.

jhicken:
The idea here is to give you an estimate of how long each spec is going to take to run. This could also be used to add heuristics to a run. But that is way more ambiguous. Since you could make an estimate on how long all the specs will run in tandem. Or based on having 5 test runners. Or based on all in parallel. This needs more thought. I think I good guess most of the time would be getting the earliest spec start time and latest spec end time for the last 5 runs and averaging that time gap but 🤷  it does not cover everycase.

Screen Shot 2020-09-01 at 10 20 00 AM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
   },
 
   rules: {
+    '@typescript-eslint/no-unused-vars': 'error',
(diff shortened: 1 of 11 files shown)


agoldis (approved):
Great job, as always @jhicken! I have added some clean-up and formatting on top of your work.

PR #129, opened 2020-09-02
Add projects

jhicken:
This is build on top of #127 
So if or when we merge this that will need to be merged first to understand some of this.

However, this will enable us to have configuration at the project level. I need this to implement things like github hooks. The projects list page is still a big WIP but. This is an effort to limit the size of some of the other stuff that im working on.

Screen Shot 2020-09-02 at 5 19 45 PM
Screen Shot 2020-09-02 at 5 20 40 PM

Projects are auto created when runs are created in director. And for migrations if your projectId matches what it was before this change, the project should just be created for you on the next run of your cypress test suite.

I was thinking of implementing a regex to only allow certain characters in projectIds. All alpha numerics, dashes, and underscores. Let me know your thoughts.

Next things I would like to do is add a project creation form. It will be needed for webhook urls. And make the project list page use a table/card or something other than flat links on the page.


agoldis:
That's super cool stuff ⭐️ 

-  I have pushed a commit with conflicts resolved. 
- I will need more time during the weekend to review the PR - it has quite a lot of code, but I was able to run it locally and it works
- I am thinking of creating a `next` branch that will include your incremental change. It is still a bit raw to release as is. Are you ok with that?

jhicken:
Yeah that works good for me. Do you want me to do just my next few "project" related PR's against the "next" branch. Or is this a pattern we want to follow going forward. That we merge to the "next" branch and then merge "next" into "master" at some sort of caidence?

agoldis:
> Yeah that works good for me. Do you want me to do just my next few "project" related PR's against the "next" branch. Or is this a pattern we want to follow going forward. That we merge to the "next" branch and then merge "next" into "master" at some sort of caidence?

Usually I release after each PR. If there will be more major changes like yours, we'll start using "next".

StefanS-O:
@jhicken Great stuff!

PR #136, opened 2020-09-15
Adding the UI to list and create / update projects.

jhicken:
I intentionally did not handle pagination. I think we need a project wide solution. Although it will only probably be used by projects and runs right now. 🤷  @agoldis let me know what you wanna do with pagination.

So net new stuff is.
Project list page is actually not garbage. (I can say this cause I made it originally.) Its just missing pagination.
Projects can be added.
Projects can be deleted. (and cleaned up including runs and instances) It would be nice to have it clean up S3 aswell
Projects can be updated. (disabled in the UI until there is more project data to update)
Project list page UI has a new look.
Project edit page has been added.

Screen Shot 2020-09-15 at 5 20 37 PM
Screen Shot 2020-09-15 at 5 19 20 PM

### DIFF
--- a/packages/api/src/resolvers/index.ts
+++ b/packages/api/src/resolvers/index.ts
@@ -1,5 +1,11 @@
+import { DataSource } from 'apollo-datasource';
 import { GraphQLDateTime } from 'graphql-iso-date';
 import { AppDatasources } from '@src/datasources/types';
+import { type } from 'os';
+
+type Project = {
+  projectId: string,
+}
 
 export const resolvers = {
   DateTime: GraphQLDateTime,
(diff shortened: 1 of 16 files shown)


agoldis:
That's amazing @jhicken. Looks like that is something we can release, I will review by the eo weekend. Thanks for your amazing work!

agoldis:
For pagination - I think the current implementation is on overkill, simple offset based pagination would be more than enough. I started to change it as part of auto-refresh task, but never had time to complete.

packages/api/src/resolvers/index.ts
```diff
@@ -40,14 +49,7 @@ export const resolvers = {
...
     ) => {
-      const instancesDeleteResponse = await dataSources.instancesAPI.deleteInstancesByRunIds(
-        [runId]
-      );
-      if (instancesDeleteResponse.success) {
-        return dataSources.runsAPI.deleteRunsByIds([runId]);
-      } else {
-        return instancesDeleteResponse;
-      }
+      return resolvers.Mutation.deleteRuns(_ ,{runIds:[runId]} ,{dataSources})
```
agoldis, line 52: 👍🏻

gmkumar08:
@agoldis Do the docker image: agoldis/sorry-cypress-dashboard:latest has this feature?

agoldis:
@gmkumar08 sure!

PR #146, opened 2020-09-22
Feature request issue template

coreyshirk:
Feature request issue template for our users. Figured we see how this works then we can make one for reporting bugs too.


coreyshirk:
@agoldis let me know if I'm missing anything or should make edits

Issue #144, opened 2020-09-22
Projects


agoldis:
@jhicken That's a ticket to track your effort on projects. From what I remember there're few leftovers: 

- [x] ~Fully functional project edit~
- [x] ~[Suggestion] Project human-friendly name~

agoldis:
Thinking more about projects - this might be quite a big change for existing users.
One use case I am worried about is users running tests  that assign random projectIds.... it will be a mess.

Another example is the demo deployment on Heroku - random people run tests on it, every test has a different projectId.

Do you think we can create some kind of project-less setup?

jhicken:
Yeah I'm glad your bringing this stuff up.

"One use case I am worried about is users running tests that assign random projectIds.... it will be a mess."

My question would be us this a realistic thing? It might be and my use cases just don't match it. Since your projectId is in your cypress.json it seems pretty static most of the time. Is cypress dashboard allowing for a dynamic projecteId?

The heroku case is kinda succy. And makes me think maybe we need to introduce some sort of list page search box to filter the projects list page.


"Do you think we can create some kind of project-less setup?"

Yeah... we could... It wouldn't be to difficult to implement an env var to disable projects or a special "all" project that shows all runs in the system. But i think I would instead vote for implementing a search feature as an alternative. I think that would get users where they want to be quicker in general. And continue to give them a more limited view of runs that are probably more relevant.

jhicken:
A couple other things.

With project edit I was not planning on allowing users to rename projects. Seems like an edge case where that would be helpful. And it might be a pain to implement. 

However the hooks stuff that I'm currently working on should be editable on a project at anytime. 

Is there something else you were thinking of with "Fully functional project edit"?

Can you explain what you are thinking with this a little bit more? "[Suggestion] Project human-friendly name"

agoldis:
Yeah, I am not sure if users are actually use as many different projects. It might be a super-edge case.

 > Is there something else you were thinking of with "Fully functional project edit"?
No, it jus wasn't clear what's the purpose of project edit if there's just project id and it's not editable. I didn't know you're planning to add more stuff in there. May be changing the icon / button to "configure" would be it more clear.

> Can you explain what you are thinking with this a little bit more? "[Suggestion] Project human-friendly name"

Yeah, I see "yc4ci4" for my project - it's been used in breadcrumbs and project name. Have an extra field on project entity like "Project name" would allow to display meaningful name to users. E.g. "web", while keeping the project id different.

[...]

Issue #143, opened 2020-09-22
Hooks


agoldis:
@jhicken I need your help here. Should it be a separate issue or bound to Github / slack ticket? 

Extracted from https://github.com/sorry-cypress/sorry-cypress/issues/87

jhicken:
Good question. The hooks feature could be considered completely separate from github and slack. However its certainly a prerequisite for them.

And im kinda doing the generic hooks in tandem with the github status hooks right now. So its kinda connected to the github issue.

Issue #120, opened 2020-08-05
Feature Request: Slack integration


gs-scooter:
For starters, I would like to congratulate you for developing a really great product. Setup was straightforward and it worked like a charm.

One of the things I like about Cypress' dashboard is the Slack integration and the ability to push test failure notifications to a channel. Would something like that be possible for this project?

agoldis:
Thanks for suggesting that feature! 
I like the idea! Let's see if more people want to have this kind of integration.

mwren-mshanken:
Hey there, this would certainly be useful!

I'm currently using the slack-api to roll my own integration. I run my tests via AWS codeBuild using the Cypress API:

This line triggers my cypress tests in codeBuild:

`- node cypress/scripts/deploy.js --configFile deploy-cypress.json --record --ciBuildId WS-${IMAGE_TAG}-${CODEBUILD_BUILD_NUMBER}-${CODEBUILD_START_TIME} --group ws-${IMAGE_TAG} --slackToken ${SLACK_BOT_TOKEN}`

My deploy.js script looks similar to others from the tutorials here, but I added my own Slack messaging using [Slack Layout Blocks](https://api.slack.com/reference/block-kit/blocks):

```
const cypress = require('cypress')
const yargs = require('yargs')
var Slack = require('slack')

//Set up our command line arguments
const argv = yargs.options({
  'record': {
      describe: 'Whether to record the test run',
      type: 'boolean',
      default: false
  },
  'key': {
      describe: 'Specify your secret record key',
      type: 'string',
      default: undefined
  },
  'group': {
      describe: 'Group recorded tests together under a single run name',
      type: 'string',
      default: undefined
  },
  'ciBuildId': {
      describe: 'Specify a unique identifier for a run to enable grouping or parallelization',
      type: 'string',
      default: undefined
  },
  'configFile': {
    describe: 'Specify a configuration file',
    type: 'string',
    default: undefined
  },
  'slackToken': {
    describe: 'Connect to slack',
    type: 'string',
    default: undefined
  }
}).help()
.argv

// Allow use of environment variables as well
argv.record = 'CYPRESS_RECORD' in process.env ? process.env.CYPRESS_RECORD : argv.record;
argv.key = 'CYPRESS_RECORD_KEY' in process.env ? process.env.CYPRESS_RECORD_KEY : argv.key;
[...]

Issue #14, opened 2019-11-12
Github Integration


IvanBioRender:
It would be nice to have a Github integration that reports test results as status checks.

agoldis:
Hey @IvanBioRender, usually the CI environment where you are runnin the tests provides status updates for github / version control provider.

Any chance you can elaborate on your use cases so I can get more familiar with requirements?

IvanBioRender:
Hey Andrew, ideally it would work the same way as Cypress does it:
https://docs.cypress.io/guides/dashboard/github-integration.html

The dashboard would have a Github token (or be installed as a Github app
into your repo), allowing it to use OctoKit to access the Github API. Once
it finishes a test run, it would report a status check for the
corresponding pull request, with a summary of the results (or a link to the
full report on the dashboard).

Thanks for this project btw, it's super useful! I don't have much time
right now, but I'd love to contribute in the future.

agoldis:
Gotcha! Will give it a try

agoldis:
After the initial exploration both Github Apps and Github OAuth integrations assumes an existence of integration "owner".

E.g. github app would be assigned a secret key and `appId`. Same for OAuth integration - it require creation of a distinct OAuth app for each `sorry-cypress` installation.

The on-premise nature of the project complicates the solution. 

May be sometime in future we can have a cloud-based cypress dashboard that can own those integrations, and allow access for different users. This model is problematic though - it is expensive (who'd pay for that?) and even if someone pays, it is kind of unethical - it would be a direct competition with cypress team.

For now I would suggest to keep this feature in backlog and try reconsider it in future.

IvanBioRender:
You make a very good point! I've set up Github apps for personal projects,
and having to do it yourself is really tedious.

I think it's with well-written documentation on the setup it wouldn't be
too bad - I may try hacking this onto sorry-cypress at some point, if I do
I'll update you :)

Thanks again for the fast response to this request!

agoldis:
@jhicken is you what you're trying to achieve with projects and hooks?

jhicken:
Yarp.

I'm pretty close aswell. I'm introducing a couple kinds of hooks that you will define on you project. One of them is a github status set of hooks that will post the status of your tests to your github prs on run start, run finish, instance start, and instance finish.

jhicken:
@IvanBioRender 
Have you tested this integration to see if it covers your use case?

denis-domanskii:
[...]

PR #148, opened 2020-09-28
Project hooks  and github status hooks on cypress v5

jhicken:
So this relates to...
#144 
#14 
#143 

And has implications for #120 

This introduces a new set of forms on the project edit view that allow you to setup a few types of hooks.

Webhooks: These are pretty straight forward. Every time one of the event hooks are called, json representations of the run and instance involved in the event plus some extra stats data will be sent in a POST call from sorry-cypress to url that is defined. You can also provide headers that will be passed in the post call. 

Github Status Hooks: Every time a run starts, finishes or an instance finishes. The provided github repository will receive a status update with the current results and a link to the run. Currently a github personal access token with status access on the repo is required.

Screen Shot 2020-09-27 at 9 00 22 PM

github status
Screen Shot 2020-09-27 at 8 56 28 PM

This is currently for v5 but I have a non v5 version of the branch as well if we need it.


agoldis:
That's 💥 will get it out soon!

packages/director/package.json
```diff
@@ -24,7 +24,10 @@
     "clean": "rimraf ./dist"
   },
   "dependencies": {
+    "@types/minio": "^7.0.5",
```
agoldis, line 27: Please move to dev

agoldis, line 27: This is a dev dep

agoldis (commented):
This is super super great 🚀💥⚡️. Thanks a lot for your great ideas and work! 

- I have changed the base for that PR to `next`
- I have pushed few changes to `next` to put some focus on dev experience and code quality - eslint, typescript errors and prettier. Now there are more people working on that I want to keep the project clean and avoid broken windows.

As for the code / implementation:

- Let's try to keep `api` and `director` typed, I didn't do a great job with types on dashboard and have been suffering. Please tell me is that's too much effort, I will try to help
- Did you consider a pub/sub for communicating events in director? I am concerned by putting too much logic into routing functions.
- I noticed few opportunities to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

Also, see inline comments

packages/api/src/datasources/projects.ts
```diff
@@ -19,6 +21,56 @@ const getSortByAggregation = (direction = 'DESC') => ({
...
+  const oldProject = await getProjectById(updatedProject.projectId);
+
+  // This is to ensure that we keep github tokens when the user only updaing the url
+  if (updatedProject && updatedProject.hooks) {
+    updatedProject.hooks = updatedProject.hooks.map((hook) => {
+      if (!hook.githubToken) {
+        const oldhook =
+          oldProject &&
+          oldProject.hooks &&
+          oldProject.hooks.find((oldHook) => oldHook.hookId === hook.hookId);
```
agoldis, line 63:
```suggestion
      if (!hook.githubToken) {
        const oldhook = oldProject.?hooks?.find((oldHook) => oldHook.hookId === hook.hookId);
```

packages/director/src/app.ts
```diff
@@ -27,18 +30,29 @@ app.get('/', (_, res) =>
...
+    return res
+      .status(403)
+      .send(`Provided record key '${recordKey}' is not allowed`);
   }
 
   console.log(`>> Machine is joining a run`, { ciBuildId });
 
-  const response = await app.get('executionDriver').createRun(req.body);
+  const response = await executionDriver.createRun(req.body);
+  const runWithSpecs = await executionDriver.getRunWithSpecs(response.runId);
```
agoldis, line 48: I would say that's a classic use case for a pub/sub

```ts
pub.emit(hookEvents.RUN_START, { runId: response.runId });
```

packages/director/src/app.ts
```diff
@@ -95,6 +124,49 @@ app.put(
...
+          results: any;
+        },
+        index: number
+      ) => {
+        return (
+          !currentSpec.claimed || !run.specsFull[index].results || wasRunning
+        );
+      },
+      false
+    );
```
agoldis, line 144: Should it be a distinct method?

packages/director/src/app.ts
```diff
@@ -95,6 +124,49 @@ app.put(
...
+          reportData: {
+            run,
+            instance,
+          },
+          project,
+        });
+        console.log(`<< RUN_FINISH hook called`, run.runId);
+      }
+    });
+
```
agoldis, line 169: This new logic seems like a side effect - another argument for using some kind of emitter

packages/director/src/execution/mongo/runs/run.model.ts
```diff
@@ -3,15 +3,74 @@ import { getMongoDB } from '@src/lib/mongo';
...
+      await getMongoDB()
+        .collection('runs')
+        .aggregate([
+          matchRunAggregation(id),
+          projectAggregation,
+          lookupAggregation,
+        ])
+        .toArray()
+    ).pop()
+  );
```
agoldis, line 63: looking at this probably we'll need a proper solution for code sharing, no action required now though

packages/director/src/execution/mongo/runs/run.model.ts
```diff
@@ -3,15 +3,74 @@ import { getMongoDB } from '@src/lib/mongo';
...
   await getMongoDB().collection('runs').findOne({ runId: id });
 
 export const createRun = async (run: Run) => {
   try {
-    const { result } = await getMongoDB()
+    const result = await getMongoDB()
       .collection('runs')
       .insertOne(getSanitizedMongoObject(run));
-    return result;
+    return result.ops[0];
```
agoldis, line 73: Why did it change? couldn't follow

packages/director/src/lib/hooksReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+              );
+            });
+          }
+        }
+      });
+  } catch (error) {
+    console.error(`Failed to run hook at for ${project.projectId}`, error);
+  }
+  return Promise.resolve();
+}
```
agoldis, line 207: I hope that works :) Breaking that down would improve readability and review-ability

packages/director/src/lib/hooksReporter.ts
```diff
@@ -0,0 +1,207 @@
...
+import axios from 'axios';
+import { getDashboardRunURL } from '@src/lib/urls';
+import { Instance } from '@src/types/instance.types';
+import { Run } from '@src/types/run.types';
+import { Project, Hook } from '@src/types/project.types';
+import { hookReportSchema } from '@src/lib/schemas';
+import Ajv from 'ajv';
+
+const ajv = new Ajv({ removeAdditional: 'all' });
+const cleanHookReportData = ajv.compile(hookReportSchema);
```
agoldis, line 11: Interesting, never used before 👀

packages/api/src/datasources/projects.ts
jhicken: Totally. I would have but ... optional chaining is only setup on dashboard. I attempted to implement it on director and typescript pooped the bed. So kinda backed off it for now. If you wanna get it functional I would be super happy to updated all these cause optional chaining is way more readable.

packages/director/package.json
jhicken: Im pretty sure this was an artifact of a rebase I did on the v5 branch. Ill move it.

packages/director/src/app.ts
jhicken: This is probably personal opinion. In my early days of programming I delt with the event emmiter pattern a lot and grew a major distaste for it. Particularly when stuff goes wrong and your emitter is hooked up to multiple subscribers. Tracking down subscriber order is a massive pain. Also just putting break points and trying to debug into a subscriber is kinda succy. Since event emitters push things to the back of the call stack.

If your not using the built in event system its not as big of a deal. Cause you can follow the code flow a little better. But I generally shy away from it cause I find it more understandable to track direct function calls.

That said if you would like me to change it I still can.

packages/director/src/app.ts
jhicken: What alternative are you thinking of? 

I separated it into a function for readability. If thats what you are asking about.

packages/director/src/app.ts
jhicken: Maybe you can explain what your thinking a little more here.

Side effect is one of the newer hipster terms that tends to be used as a ... "this thing runs async code and sets state somewhere"
If thats what your meaning then... yes this runs async code. If not please correct me.

What about an emitter would make this better? 

Sorry if this comes off as abrasive. That not what im going for I just want to understand. And get on the same page.

packages/director/src/execution/mongo/runs/run.model.ts
jhicken: Yup I actually meant to comment about this. I attempted lots of stuff. I tried to copy in all the src folders into the dev and prod containers and setup symlinks to @sorry-cypress/<package-name> for all the projects. That blew up due to missing dependency chains. So I took a different approach and I copied the same structure of file hierarchy into the docker containers. I copied all the package manifest stuff first. Then ran `yarn` then tried to import dependancies. With this approach I had some weird inconsistency in the typescript configuration across <packages>. I tried one more thing but Im blanking on it right now. After wasting a couple days on it I backed off and did this. But I think a better solution may be to just have a separate <package> that holds all the shared files and have all the other <pacakges> use it as a dependency in thier package.json's.


I learned the node docker images have yarn. Who knew?

packages/director/src/execution/mongo/runs/run.model.ts
jhicken: The original result is the mongo operation results. Which is not what I was expecting to get back when I told the execution driver to create a run. So I modified the result to return the actual created run from the mongo operation. Which in the case happend to be the result of the 0ith operation.

This is where the result is used.
https://github.com/sorry-cypress/sorry-cypress/pull/148/files#diff-642ad9f68f8d922e3064843951f75474R47

packages/api/src/datasources/projects.ts
agoldis: I am pretty sure it is working now

packages/director/src/lib/hooksReporter.ts
jhicken: This is kinda a weird case. Typescript is kinda succy when you wanna create an object on the fly inside a function. So I used a schema validator to remove the un wanted/needed fields from an object with an already defined type. The main reason this was needed was the data set being sent back to the webhooks was too large. Node was flipping out about the size.

Honestly I think using plain javascript to build the object on the fly is cleaner but again typescript makes that dang near impossible.

packages/director/src/lib/hooksReporter.ts
jhicken: optional chaining would help :)

How are you thinking to break it down? Like should I put the generic webhook code in its own function?

Next to that I commented on what I thought was confusing. What did I miss that I can make less succy?

packages/director/src/app.ts
agoldis: I agree that it can go left when abused - there are pros and cons. I did see emitters working nicely when controlled and used strictly. I also suffered a lot while working on systems that abused the pattern and lost control of emitters / listeners. 

Bottom line, if you dislike emitters so much, let's try to find a different solution. What I am looking for is to separate the functionality that deals with director primary responsibility (manage the tests) from hooks functionality.

Simply extracting all hooks functionality from `app.ts` into a separate module would also work for me.

packages/director/src/app.ts
agoldis: Yes, this

packages/director/src/app.ts
agoldis: Yeah, sorry for dropping buzzwords. I just wanted to separate hooks from the rest of functionality in `app.ts`

> Sorry if this comes off as abrasive. 

Code reviews are a subtle art :-) You are doing great 👍🏻

packages/director/src/execution/mongo/runs/run.model.ts
agoldis: Yeah, seems like dockerizing the packages prevents using yarn workspaces properly. I will have to take a deeper look, it's an interesting problem.

packages/api/src/datasources/projects.ts
jhicken: It is! Huzzah! I just fixed these all locally.

jhicken:
I replied to everything inline.

> Let's try to keep api and director typed, I didn't do a great job with types on dashboard and have been suffering. Please tell me is that's too much effort, I will try to help

Thanks! Yeah help on that would be super appreciated. Sorry if im a pain on that. I'm not a big typescript user. Im happy to make whatever changes you think we need to. So if there are places you want me to change types just call them out. Also there seems to be a lot of ways and places to define said types. Would you like them in a single file? Do you want them in the source file that is most closely related to them. Do you want them inline on functions themselfs. Do you want them in a file separate file with the same name as the file that your working on. Do we wanna be sharing them across `<packages>`? Sorry part of why its hard for me is that I dont think we have it nailed down when and where types are created and stored.

Guide me please.

packages/director/src/lib/hooksReporter.ts
agoldis: ok, nvm it's more readable at 11am comparing to 1am

packages/director/src/app.ts
jhicken: > Code reviews are a subtle art

Well said! :)

agoldis:
> So if there are places you want me to change types just call them out. Also there seems to be a lot of ways and places to define said types. 

- Would you like them in a single file? **No**
- Do you want them in the source file that is most closely related to them. **See below**
- Do you want them inline on functions themselves. **See below**
- Do you want them in a file separate file with the same name as the file that your working on. **No** 
- Do we wanna be sharing them across <packages>? **Ideally yes, but we don't have a good solution for that now.**

--- 

I am not opinionated about that, what is important is consistency. Suggesting to mimic the current [director](https://github.com/sorry-cypress/sorry-cypress/tree/master/packages/director/src) structure.

For trivial types inline function would work ([example](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/api/src/datasources/runs.ts#L160))

For more complex types, let's have a `types` file that is scoped a specific domain, e.g. s3 screenshot driver has its own types  [example](https://github.com/sorry-cypress/sorry-cypress/tree/master/packages/director/src/screenshots/s3)

For cross-functional shared types let's use root "types" [directory](https://github.com/sorry-cypress/sorry-cypress/tree/master/packages/director/src/types). 

For generated types (from GraphQL) - `generated/...` directory

I am marking the specific functions missing types...

packages/api/src/datasources/projects.ts
```diff
@@ -1,5 +1,7 @@
+import { hookTypes } from '@src/duplicatedFromDirector/hooksEnums';
 import { getMongoDB, init } from '@src/lib/mongo';
 import { DataSource } from 'apollo-datasource';
+import uuid from 'uuid/v4';
```
agoldis, line 4:
```suggestion
import uuid from 'uuid/v4';
import { Project } from '@src/duplicatedFromDirector/project.types';
```

packages/api/src/datasources/projects.ts
```diff
@@ -19,6 +21,56 @@ const getSortByAggregation = (direction = 'DESC') => ({
   },
 });
 
+const addHookIdsToProjectHooks = (project) => {
```
agoldis, line 24:
```suggestion
const addHookIdsToProjectHooks = (project: Project) => {
```

packages/api/src/datasources/projects.ts
```diff
@@ -19,6 +21,56 @@ const getSortByAggregation = (direction = 'DESC') => ({
...
+  if (project && project.hooks) {
+    project.hooks = project.hooks.map((hook) => {
+      hook.hookId = hook.hookId || uuid();
+      return hook;
+    });
+  }
+  return project;
+};
+
+const removeUnusedHookDataFromProject = (project) => {
```
agoldis, line 34:
```suggestion
const removeUnusedHookDataFromProject = (project: Project) => {
```

packages/api/src/datasources/projects.ts
```diff
@@ -27,29 +79,37 @@ export class ProjectsAPI extends DataSource {
   async getProjectById(id: string) {
     const result = getMongoDB()
       .collection('projects')
-      .aggregate([{
+      .aggregate([
```
agoldis, line 82:
```suggestion
      .aggregate<Project>([
```

packages/api/src/duplicatedFromDirector/project.types.ts
```diff
@@ -0,0 +1,14 @@
+export type Hook = {
+  hookId: string;
+  url: string;
+  headers: string;
+  hookEvents: [string];
```
agoldis, line 5:
```suggestion
  hookEvents: string[];
```

packages/api/src/duplicatedFromDirector/project.types.ts
```diff
@@ -0,0 +1,14 @@
...
+  hookType: string;
+  githubToken: string;
+};
+
+export interface Project {
+  projectId: string;
+  createdAt: string;
+  hooks?: [Hook];
```
agoldis, line 13:
```suggestion
  hooks?: Hook[];
```

packages/dashboard/src/duplicatedFromDirector/project.types.ts
```diff
@@ -0,0 +1,14 @@
+export type Hook = {
+  hookId: string;
+  url: string;
+  headers: string;
+  hookEvents: [string];
```
agoldis, line 5:
```suggestion
  hookEvents: string[];
```

packages/dashboard/src/duplicatedFromDirector/project.types.ts
```diff
@@ -0,0 +1,14 @@
...
+  hookType: string;
+  githubToken: string;
+};
+
+export interface Project {
+  projectId: string;
+  createdAt: string;
+  hooks?: [Hook];
```
agoldis, line 13:
```suggestion
  hooks?: Hook[];
```

jhicken:
Sweet. Would you like me to make these extra changes in another PR?

agoldis:
> Sweet. Would you like me to make these extra changes in another PR?

If you already done some refactoring - sure, otherwise don't waste your time :)

PR #151, opened 2020-10-02
Add kubernetes example deployment

tico24:
I appreciate that it's not a huge amount of work, but I deployed Sorry Cypress into Kubernetes and I thought this might help others in the future.

### DIFF
--- a/kubernetes-full.yml
+++ b/kubernetes-full.yml
@@ -1,0 +1,293 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: mongo-storage-claim
+  labels:
+    app: mongo-storage-claim
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mongo
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: mongo
+  template:
+    metadata:
+      name: mongo
+      labels: 
+        app: mongo
+    spec:
+      containers:
+      - image: mongo:4.0
+        imagePullPolicy: "Always"
+        name: mongo
+        ports:
+        - containerPort: 27017
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 27017
+          periodSeconds: 10
+          timeoutSeconds: 5
+          successThreshold: 2
+          failureThreshold: 5
+        resources:
+          requests:
+            memory: "128M"
+            cpu: 0.2
+          limits:
+            memory: "512M"
+            cpu: 0.5
+        volumeMounts:
+          - name: mongo-storage
+            mountPath: /data/db
+      restartPolicy: Always
+      serviceAccountName: ""
+      volumes:
+        - name: mongo-storage
+          persistentVolumeClaim:
+            claimName: mongo-storage-claim
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: api
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: api
+  template:
+    metadata:
+      labels:
+        app: api
+      name: api
+    spec:
+      containers:
+      - env:
+        - name: MONGODB_DATABASE
+          value: sorry-cypress
+        - name: MONGODB_URI
+          value: mongodb://mongo-service:27017
+        image: agoldis/sorry-cypress-api:cypress-v5
+        imagePullPolicy: "Always"
+        name: api
+        ports:
+        - containerPort: 4000
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 4000
+          periodSeconds: 10
+          timeoutSeconds: 5
+          successThreshold: 2
+          failureThreshold: 5
+        resources:
+          requests:
+            memory: "128M"
+            cpu: 0.2
+          limits:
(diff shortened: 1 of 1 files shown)


kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+          serviceName: director-service
+          servicePort: 1234
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cypress-s3-secrets
+data:
+  AWS_SECRET_ACCESS_KEY: VGhpcyBpcyB0aGUgY3J1bWJob2xl
+  AWS_ACCESS_KEY_ID: Tm8gcGVla2luZw==
```
agoldis, line 293: 👀

kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+    metadata:
+      labels:
+        app: dashboard
+      name: dashboard
+    spec:
+      containers:
+      - env:
+        - name: GRAPHQL_SCHEMA_URL
+          value: [YOUR-API-URL-HERE]
+        image: agoldis/sorry-cypress-dashboard:cypress-v5
```
agoldis, line 190: I have just merged cypress v5 into master
```suggestion
        image: agoldis/sorry-cypress-dashboard:latest
```

kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+        app: api
+      name: api
+    spec:
+      containers:
+      - env:
+        - name: MONGODB_DATABASE
+          value: sorry-cypress
+        - name: MONGODB_URI
+          value: mongodb://mongo-service:27017
+        image: agoldis/sorry-cypress-api:cypress-v5
```
agoldis, line 82:
```suggestion
        image: agoldis/sorry-cypress-api:latest
```

kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+          valueFrom:
+              secretKeyRef:
+                name: cypress-s3-secrets
+                key: AWS_ACCESS_KEY_ID
+        - name: AWS_SECRET_ACCESS_KEY
+          valueFrom:
+              secretKeyRef:
+                name: cypress-s3-secrets
+                key: AWS_SECRET_ACCESS_KEY
+        image: agoldis/sorry-cypress-director:cypress-v5
```
agoldis, line 147:
```suggestion
        image: agoldis/sorry-cypress-director:latest
```

kubernetes-full.yml
tico24: Placeholders, don't worry :)

agoldis:
That's awesome! Why not huge, it's very helpful, i think it will be quite an enabler.
Any chance you write few paragraphs of instruction - I can use it for documentation and for myselft

tico24:
I am half-way through writing a blog if it helps?
https://crumbhole.com/playing-with-sorry-cypress-and-kubernetes/

kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+      - backend:
+          serviceName: dashboard-service
+          servicePort: 8080
+  - host: [YOUR-API-URL-HERE]
+    http:
+      paths:
+      - backend:
+          serviceName: api-service
+          servicePort: 4000
+  - host: [YOUR-DIRECTOR-URL-HERE]
```
agoldis, line 280: Sorry if I am stupid, does it allow to access "internal" services using the URLs in `[]`?

kubernetes-full.yml
```diff
@@ -0,0 +1,293 @@
...
+  template:
+    metadata:
+      labels:
+        app: dashboard
+      name: dashboard
+    spec:
+      containers:
+      - env:
+        - name: GRAPHQL_SCHEMA_URL
+          value: [YOUR-API-URL-HERE]
```
agoldis, line 189: Is it possible to use some kind of "internal DNS" to route traffic to API service? I have 0 kubernetes experience

kubernetes-full.yml
tico24: This is essentially what creates the url, provisioning a loadbalancer and even possibly the DNS entries (if you have k8s set up to do it).

agoldis:
I will run this soon, right in the middle of a release - awesome work ! 🎉

kubernetes-full.yml
tico24: Yes, that's possible. I actually already do that on line 130. The director pod accesses mongo internally.
If that's what's supposed to happen here then I'll adjust.

tico24:
Actually, I've just spotted a problem with the api readiness probe. Hold on and I'll push a revised yaml.

kubernetes-full.yml
agoldis: Yes, `dashboard` pull graphql schema definition from the API service

tico24: So it looks like the dashboard also relies on this URL when the end user visits the dashboad through a browser.
Changing this to the internally-accessible dns entry prevents the end user from loading their data.

So I believe the right thing to do in this instance is indeed to call the external API URL.

agoldis: Oh right, it need to access the API from the user-space

agoldis:
@all-contributors please add @tico24 for Content and example

agoldis:
@all-contributors please add @tico24 for content and example

agoldis (approved):

PR #178, opened 2020-10-14
Group parameter usage while generating runId

Upgreydd:


Upgreydd:
Specifying the same runId forces agents to use the same specs.

PR #181, opened 2020-10-15
Parsing of `--group` parameter and nesting runs into accordion

Upgreydd:
thx to @pleszkowicz

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,6 @@ credentials
 
 # Minio
 /data
+
+# IntelliJ
+.idea
--- a/packages/api/src/generated/graphql.ts
+++ b/packages/api/src/generated/graphql.ts
@@ -172,6 +172,7 @@ export type Commit = {
 export type RunMeta = {
   __typename?: 'RunMeta';
   groupId?: Maybe<Scalars['String']>;
+  group?: Maybe<Scalars['String']>;
   ciBuildId?: Maybe<Scalars['String']>;
   projectId?: Maybe<Scalars['String']>;
   commit?: Maybe<Commit>;
@@ -539,6 +540,7 @@ export type CommitResolvers<ContextType = any, ParentType extends ResolversParen
 
 export type RunMetaResolvers<ContextType = any, ParentType extends ResolversParentTypes['RunMeta'] = ResolversParentTypes['RunMeta']> = {
   groupId?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
+  group?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
   ciBuildId?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
   projectId?: Resolver<Maybe<ResolversTypes['String']>, ParentType, ContextType>;
   commit?: Resolver<Maybe<ResolversTypes['Commit']>, ParentType, ContextType>;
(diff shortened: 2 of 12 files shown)


agoldis:
@Upgreydd thanks a lot for creating the PR! 

Looking at how the original cypress designed their dashboard, runs contains groups: 

Screen Shot 2020-10-15 at 9 56 18 PM

While this PR implements an inverse relation:

Screen Shot 2020-10-15 at 9 55 36 PM

So what does it take to implement it right? Quite a lot :) Let's try to break it down.

--- 

Let's say we have 2 cypress agents running on different machines, agent A runs group A specs, agent B runs group B tests.
Groups A and B have different specs.

Here's what happens behind the scenes

1. First request from an agent hits sorry cypress (I call it [joining a run](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/director/src/app.ts#L35))

- figures out if there's an already existing run create a new run if none exist
- respond with `runId`, `machineId` and `groupId`, agents will send back those params with subsequent request

Now we keep in mind that:
  - when creating a new run sc [calculates a hash](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/director/src/lib/hash.ts#L5) that is based on `ciBuildId + commit.sha + projectId + specs.join(' ')`. If two machines send the same parameters, they will be part of the same run. To support the scenario above we need to get rid of specs list while building that hash.
  - it tries to create a new run with `runId`, if [creation fails](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/director/src/execution/mongo/runs/run.controller.ts#L48) - it means there's already an existing run, so we are returning an pre-existing run
  - to support groups, we need to change run schema to have "groups" instead of 'group' and update this field when a new group is detected for the same run
  - we also need to set group for each spec - that's in order to properly show grouping in UI and actually support parallelization on group level (more on that later)

2. Subsequent request is to ["get a task to run"](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/director/src/app.ts#L66)

An agent asks director what should be the next task to run. 

- the agents sends `runId`, `machineId` and `groupId` received from previous step. That's how parallelization works - director directs agents what's next tasks to run. 
- director identifies the next available task for a run, marks it as `claimed` and send it back to requesting agent.
- there's a subtle chance of race condition, so there's some ["advanced 👹"](https://github.com/sorry-cypress/sorry-cypress/blob/672b3ebe570e2d0e5cbab14062e4105195afa1ab/packages/director/src/execution/mongo/runs/run.model.ts#L85) DB query for that
- to support groups, we'd scope the set of "available tasks" only to the requested `groupId` 

3. API and dashboard adjustments 

Due to the changes to data structure, there are few changes required in how GraphQL resolves the data and how UI fetches the data. I think you've mostly done that part.

---

Now you see why isn't that feature ready yet. So far there were only 2 requests to support grouping, so I have been deferring this task. 

May be there's some easier and more creative solution. Let me know what do you think about this write up :)

Upgreydd:
@agoldis can I reach you online at some communicator to talk about this for few minutes. I'll try to give you little insight

agoldis:
Sure, let's try to sync on [slack](https://app.slack.com/client/T013PH329PT)

Upgreydd:
@agoldis Sure but it's asking me to provide workspace name, login, and password. I've tried to provide workspace "sorry-cypress" but probably I'm not authorized to join without an invitation link or existing account.

agoldis:
My bad
https://join.slack.com/t/sorry-cypress/shared_invite/zt-eis1h6jl-tJELaD7q9UGEhMP8WHJOaw

Upgreydd:
@agoldis check slack please :+1:

Upgreydd:
Resolved in https://github.com/sorry-cypress/sorry-cypress/pull/189

PR #184, opened 2020-10-16
Group by ciBuildId

Upgreydd:
thx to @pleszkowicz

- Adds Toggle for stacking builds by `ciBuildId`
- Nice formatting and minimizing would be done in next PR

Stack runs disabled:
image
Stack runs enabled:
image
Stack runs enabled and accordion expanded:
image

### DIFF
--- a/packages/dashboard/src/views/RunsView.tsx
+++ b/packages/dashboard/src/views/RunsView.tsx
@@ -1,8 +1,9 @@
 import { getProjectPath, navStructure } from '@src/lib/navigation';
 import { Button } from 'bold-ui';
 import React, { useLayoutEffect } from 'react';
-import { RunSummary } from '../components/run/summary';
-import { useGetRunsFeedQuery } from '../generated/graphql';
+import { RunSummary, RunCiGroupSummary } from '../components/run/summary';
+import { FullRunSpec, Run, useGetRunsFeedQuery } from '../generated/graphql';
+import { useStackRuns } from '@src/hooks/useStackRuns';
 
 type RunsViewProps = {
   match: {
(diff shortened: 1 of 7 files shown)


packages/dashboard/src/views/RunsView.tsx
```diff
@@ -73,11 +75,30 @@ export function RunsView({
   if (!runFeed.runs.length) {
     return <div>No runs have started on this project.</div>;
   }
+
+  const groupedRuns: any[] = [];
+  data.runFeed.runs.forEach((run) => {
```
jhicken, line 80: What do you do if all the runs for this group are not in this page of results?

jhicken: I think this needs more thought. Particularly if we add filtering to these run results. Which I think we should.

agoldis:
"Run" is actually a "group by ciBuildId", I think removing specs from runId hash would solve the problem for @Upgreydd. Will validate that nothing goes left tonight. More details in https://github.com/sorry-cypress/sorry-cypress/pull/181. 

--- 

- when creating a new run sc calculates a hash that is based on `ciBuildId + commit.sha + projectId + specs.join(' ')`. If two machines send identical parameters, they will be associated with the same run. To support the scenario above we need to get rid of specs list while building that hash.

- it tries to create a new run with runId, if creation fails - there's already an existing run, so we are returning this pre-existing run

packages/dashboard/src/views/RunsView.tsx
Upgreydd: it will be updated after clicking `Load more`

Upgreydd:
Resolved in https://github.com/sorry-cypress/sorry-cypress/pull/189

luengnat:
@Upgreydd  the accordion is still not in the build. Should we add the support for that as well?

PR #192, opened 2020-10-22
Remove unneeded port exposure in docker-compose files

### LINKED ISSUE #190: Reconsider docker-compose configuration
The current docker-compose files expose all ports to the host machine, which seems very odd to me. Combined with no authentication configuration for Mongo DB [docker-compose.full.yml](https://github.com/sorry-cypress/sorry-cypress/blob/master/docker-compose.full.yml#L7), it is just exposed to the outside and anyone can access it. There are tons of bots constantly scanning for open databases and you can see in #174 that these servers are actually hacked. 

When using docker-compose all containers have access to each other without having to expose the ports, so I would remove the exposed ports of mongo DB.

```
  mongo:
    image: mongo:4.0
```

Any ideas @agoldis ?

dlavrenuek:
Closes #190 

This PR addresses exposure of unneeded ports in docker-compose files. These files are very convenient to use but might provide a high risk when used for running on an actual server/cloud. Because the [documentation](https://sorry-cypress.dev/quickstart) only mentions running the `docker-compose.minio.yml`, I would like to make the intend more clear by adding a `docker-compose.dev.yml` that can be used on top of other docker compose files to expose extra ports.

I feel like these changes could be documented somewhere but on the other hand anyone who works on the project and needs the extra ports locally will figure it out, since this was not documented before either.

### Changes included
- `docker-compose.yml`
  - removed node debug port `9230` from `api` service
  - removed node debug port `9229` from `director` service
  - removed MongoDB connection port `27017` from `mongo` service
- `docker-compose.full.yml`
  - removed MongoDB connection port `27017` from `mongo` service
- `docker-compose.minio.yml`
  - removed MongoDB connection port `27017` from `mongo` service

### Running the services
Starting the services with exposed ports as they were before the change can be done by including the `docker-compose.dev.yml` file as following:

#### default
```bash
docker-compose -f docker-compose.yml -f docker-compose.dev.yml up
```

#### minio
```bash
docker-compose -f docker-compose.minio.yml -f docker-compose.dev.yml up
```

#### full
```bash
docker-compose -f docker-compose.full.yml -f docker-compose.dev.yml up
```

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -1,0 +1,15 @@
+## Running the services in dev mode
+
+Starting the services with exposed ports by including the `docker-compose.dev.yml` file as following:
+
+### default
+
+`docker-compose -f docker-compose.yml -f docker-compose.dev.yml up`
+
+### minio
+
+`docker-compose -f docker-compose.minio.yml -f docker-compose.dev.yml up`
+
+### full
+
+`docker-compose -f docker-compose.full.yml -f docker-compose.dev.yml up`
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,0 +1,14 @@
+version: '3.6'
+
+services:
+  mongo:
+    ports:
+      - 27017:27017
+
+  director:
+    ports:
+      - 9229:9229
+
+  api:
+    ports:
+      - 9230:9230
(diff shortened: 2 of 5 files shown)


agoldis:
Thanks for taking care of that issue!

agoldis (approved):

Issue #193, opened 2020-10-26
[Bug] Hooks don't work


gespispace:
## Summary

Sorry cypress does not send hook, but should send hook

## How to reproduce
I don't know, but it is error
```
director_1       | (node:24) UnhandledPromiseRejectionWarning: TypeError: Cannot create property 'hookEvent' on boolean 'true'
director_1       |     at reportToGenericWebHook (/app/dist/lib/hooksReporter.js:143:28)
director_1       |     at /app/dist/lib/hooksReporter.js:174:16
director_1       |     at Array.forEach (<anonymous>)
director_1       |     at reportToHook (/app/dist/lib/hooksReporter.js:166:152)
director_1       |     at handleUpdateInstance (/app/dist/api/instances.js:103:35)
director_1       |     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
and hook 
![image](https://user-images.githubusercontent.com/37763089/97125756-bec2dd00-1756-11eb-8264-24a2ee77cc1a.png)

## Environment

- sorry-cypress version: `latest`
- cypress version: `5.4.0`
- platform: `docker-compose.minio.yml`
- service: `director`

gespispace:
what other information is needed to solve the problem?

jhicken:
Odd Ill have a look at it in the morning.

gespispace:
@agoldis bug not fix. 
![image](https://user-images.githubusercontent.com/37763089/97477676-ba264080-1971-11eb-9b26-7e65241d36b4.png)

gespispace:
latest images
![image](https://user-images.githubusercontent.com/37763089/97477938-12f5d900-1972-11eb-8eb7-22396139d0f4.png)

agoldis:
kk well fix later

PR #198, opened 2020-10-28
Fix bug if header don't set. #193

gespispace:
you need to check that there are headers, and then try to parse them.

it is error 
image

### DIFF
--- a/packages/director/src/lib/hooksReporter.ts
+++ b/packages/director/src/lib/hooksReporter.ts
@@ -173,7 +173,7 @@ async function reportToGenericWebHook({
     );
     return axios({
       method: 'post',
-      headers: JSON.parse(hook.headers) || {},
+      headers: hook.headers ? JSON.parse(hook.headers) : {},
       url: hook.url,
       data: reportData,
     }).catch((err) => {


gespispace:
if I try to look through the api dashboard, I see that the `headers` is undefined and `JSON.parse(undefined)` return error
image

jhicken (approved):
Nice! Good find!

jhicken:
> you need to check that there are headers, and then try to parse them.
> 
> it is error
> image

So in this it looks like your problem is something different. It looks like you maybe had malformed json in the header field and the character `u` broke stuff. 

I think the real fix for this would be to add severside validation for api side of hooks. And validate that json much earlier. That said this pr is fine and wont hurt anything.

gespispace:
> > you need to check that there are headers, and then try to parse them.
> > it is error
> > image
> 
> So in this it looks like your problem is something different. It looks like you maybe had malformed json in the header field and the character `u` broke stuff.
> 
> I think the real fix for this would be to add severside validation for api side of hooks. And validate that json much earlier. That said this pr is fine and wont hurt anything.

But my headers was empty
image

jhicken:
By chance could you capture the network response from the network panel. and post that?

It totally possible that there is a bug here. but without more info its hard to know. I think the best thing to do would be put a debugger or a log statment right here in the code. to see what data you have stored in the db for the headers field.

Issue #240, opened 2020-12-23
Create a helm chart


fsmaia:
- [x] I have seached issues for similar feature requests

## What is this feature about (1 sentence)

This feature is about having a helm chart to automate deployment using [Helm](https://github.com/helm/helm)

## Why is it needed? What is the value? For whom do we build it?

In one of the main guides listed in the docs, the writer used a tool to create a Helm chart:
https://medium.com/@adityahbk/cypress-parallelization-on-jenkins-using-sorry-cypress-197a86ad8ed1

IMHO it would increase the popularity of this amazing tool, and it would make it easier to deploy, also by forcing some sane defaults.

## High-level feature overview

Taking lighthouse-ci and sonarqube as examples:
* https://github.com/sebastien-prudhomme/cowboysysop-charts/tree/lighthouse-ci/charts/lighthouse-ci
* https://github.com/Oteemo/charts/tree/master/charts/sonarqube

They added a boilerplate to make the repo a Helm chart repository. I haven't seen any Helm charts within a regular repository like this one, so I don't know if a packages/helm directory would be enough, or if additional boilerplate would be required.

tico24:
I'll take this. It'll probably be in its own repo (at least initially) due to how the Helm releases work.

tico24:
I've had a first stab at it. It's currently in my own repo [here](https://github.com/tico24/sorry-cypress-helm/tree/main/sorry-cypress-helm) because I'm nowhere near ready to taint this repo with it!

It's very opinionated, but seems to work with the minimal testing I've done. Would appreciate your feedback.

tico24:
I've now spent a few hours on it and its reasonably robust. I believe it's probably "good enough" for an initial release, and then we can grow/tweak it as requirements come through. I'd still appreciate some real-world usage so let me know your thoughts @agoldis @fsmaia.

The default deployment is a non-persistent affair. This should make it easier for new people to deploy. However, persistent mongo and S3 access is available and configurable in the values.yml with examples. I believe I've included options for all the various environment variables that are documented.

@agoldis There's a couple of things I can think of that'll need some thought before we merge in:

**1. Where should it live?**
The helm chart should get released whenever the software gets a release update, but it'll also probably get released in between those periods (bug fixes, enhancements etc). I don't know your release flow, but maybe if this interferes with your releases, it should have it's own repo?

[...]

PR #242, opened 2020-12-24
#240 Create a Helm Chart

tico24:
Hopefully I've put everything in the right place. Apologies if not... it's very early here!

This is the initial attempt at a Helm chart for Sorry Cypress. By default there's no persistence, it uses the dummy screenshots driver and the inmemory db driver.

However, tweaking the values.yml file will allow the user to both create (and use) a persistent mongo instance, and configure S3 screenshot storage.


agoldis:
@tico24 Sorry for taking so long. I have been afk a lot during the holidays, I appreciate a lot what you've done! Will review soon!

agoldis (approved):

agoldis:
@tico24 @fsmaia I was sure that adding org members would grant sufficient permissions to create branches etc... apparently not. I have invited you to a team that has maintainer access to this repo, and admin for https://github.com/sorry-cypress/charts

Please tell me if that's sufficient

tico24:
ok @agoldis @fsmaia I'm now not sure why the linter is still failing. I'm reluctant to merge this in with it failing.  It passes on my downstream fork:

https://github.com/tico24/sorry-cypress/runs/1654760212?check_suite_focus=true
https://github.com/tico24/sorry-cypress/runs/1654760232?check_suite_focus=true

tico24:
Well one of the reasons was I forgot the release.yml so it never released the helm. I have addressed this now, but this hasn't fixed the issue.

agoldis:
@tico24  I am a complete noob, but I think until the files are actually merged into master, they won't be accessible?

tico24:
> @tico24 I am a complete noob, but I think until the files are actually merged into master, they won't be accessible?

This would make a lot of sense. Shall we just try a merge?

agoldis:
Hm... what is `chart-testing` supposed to do? Does it test the candidate?

tico24:
> Hm... what is `chart-testing` supposed to do? Does it test the candidate?

I would hope so, yes!

agoldis:
> > Hm... what is `chart-testing` supposed to do? Does it test the candidate?
> 
> I would hope so, yes!

In that case probably using the PR branch is correct, otherwise it will test the version on master?

agoldis:
sorry if it doesn't make sense - that's all quite new to me :)

tico24:
> sorry if it doesn't make sense - that's all quite new to me :)

No, that would make sense. It would explain why it works on my fork. I don't know how to tell it to look at the PR branch though.... frantically googling :)

tico24:
I could well be wrong, but I think it IS checking the PR code. It looks like it saying "as long as the PR is a merge to master, then run this". It then pulls the PR, merges with master and runs the tests... that's what the github action logs imply anyway.

tico24:
I'm afraid I'm out of ideas!

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,51 @@
...
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
+
+      - name: Run chart-testing (lint)
+        id: lint
+        uses: helm/chart-testing-action@v2.0.1
+        with:
+          command: ct lint
+          config: charts/sorry-cypress/lint.yaml
```
agoldis, line 42:
```suggestion
          config: charts/lint.yaml
```

charts/test.yaml
```diff
@@ -0,0 +1,8 @@
+# See https://github.com/helm/chart-testing#configuration
```
agoldis, line 1: is it really used?

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1.0.0
+        with:
+          install_local_path_provisioner: true
+        if: steps.lint.outputs.changed == 'true'
+
+      - name: Run chart-testing (install)
+        run: ct install
```
agoldis, line 51: should we use an explicit config file with `debug: true`?

agoldis:
> I'm afraid I'm out of ideas!

@fsmaia do you have any ideas?

charts/test.yaml
tico24: it would appear not. It's also not where we're failing though.

tico24:
I think I've worked it out. We have to release the chart first, which means it needs to be on master.. and then it should work, no? No matter what I do on my repo, it works.... so I'm still pretty confident its working. Worst case, we revert

agoldis:
I did something here https://github.com/sorry-cypress/sorry-cypress/pull/254 and it worked, but I have no idea why it worked

tico24:
it worked because I changed the "on: pull_request" line to "on: push".

PR #249, opened 2021-01-02
Reduce dashboard memory usage

slinstaedt:
I tried to further slim down api and director images using webpack, but failed to create a reliable build.

Nevertheless image sizes for api and director is about halved by locking dependency versions and excluding dev dependencies from the images using multistage builds. Image size for dashboard is less than a 10th of the original by only including the statically compiled browser client into a vanilla nginx image. 

Memory consumption during runtime for dashboard was reduced to almost nothing, due to nginx' resource consumption being barely noticable.

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -1,0 +1,43 @@
+version: '3.8'
+
+services:
+  mongo:
+    image: mongo:4.0
+
+  director:
+    build: packages/director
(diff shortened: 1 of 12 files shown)


agoldis:
@all-contributors please add @slinstaedt for code and Infrastructure

agoldis:
Awesome @slinstaedt thanks a lot!

Issue #230, opened 2020-12-19
Reduce dashboard memory usage


slinstaedt:
Just for curiosity i monitored the memory usage of the cypress docker setup and was a little bit surprised, seeing the dashboard using around 200-250MB, while all the other are fine with 50-60MB. From my understanding the dashboard is only serving the client application (aka static files) to the browsers. Normally one would expect this is a job, a simple nginx with 5MB usage can handle. As infrastructure costs is mostly related to memory usage for sorry-cypress I would like to ask, if there is a chance to reduce the memory usage.

Could you give some insights, what the dashboard docker service is actually doing?

I have searched for similar issues. There is only one (#46) mentioning the high memory usage, but basically dealing with another topic.

slinstaedt:
Sorry, had a duplicate post due to browser timeout.

agoldis:
You're right, the service is very simple - it's just a node server that server pre-compiled assets.
How are you measuring the memory? One random idea is that the docker image is bigger than needed because of node modules or leftovers from the build process.

slinstaedt:
Though I was talking about the runtime memory consumption, having a smaller image is also almost a nice idea ;) 

for the runtime memory I was using both classical `top`, and `docker stats`. Even though they might not be accurate, they will give you an idea especially in comparison to other processes.

Have you thought about using a multi-stage build for your images? They can potentially dramatically reduce the their size and depending on the runtime base image, you choose, also the memory usage.

agoldis:
@slinstaedt sorry about slow response, I have been afk a lot last few weeks.

slinstaedt:
Used some spare time during lockdown... feedback is welcome #249

agoldis:
I guess this was resolved by #249

PR #259, opened 2021-01-08
Allow dynamic port binding for dashboard

agoldis:
Allows dynamic port binding for dashboard via `PORT` env var. Hardcoded `8080` would not allow running on heroku, and, in general, should not be hardcoded. 

It is a follow up on @slinstaedt work in #230, which was awesome because it reduced image size significantly (and caused me to explore multi-staged Dockerfiles).

 @tico24
One less-desired side effect is that dashboard environment variables are now mandatory to run production. I have updated kubernetes templates accordingly. I guess we also need to update helm charts?

### DIFF
--- a/docker-compose.build.yml
+++ b/docker-compose.build.yml
@@ -37,6 +37,8 @@ services:
     image: agoldis/sorry-cypress-dashboard:${VERSION:-latest}
     environment:
       GRAPHQL_SCHEMA_URL: http://localhost:4000
+      PORT: 8080
+      CI_URL: ''
     ports:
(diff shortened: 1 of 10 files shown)


tico24:
@agoldis What's the target Sorry-Cypress release version? Helm tracks non-beta at the moment, so if this is going into a beta release we don't need to immediately jump to change the Helm chart.

agoldis:
That'd be a beta release, but it will update the latest tag on dockerhub and dashboard won't start w/o those env variables

tico24 (approved):

PR #265, opened 2021-01-18
Add auth support to MongoDB

fsmaia:
Although MongoDB can be achieved by embedding the auth mechanism/user/password into `MONGODB_URI`, this allows it in a more explicit way.

MongoDB with authentication is a corner case, it is only necessary when providing an external MongoDB database.

**Note: if the environment variables are not set, there aren't any behavior changes.**


agoldis (approved):
Looks good!

PR #266, opened 2021-01-19
Add support to ApolloClient credentials in the dashboard

fsmaia:
This PR adds a `GRAPHQL_CLIENT_CREDENTIALS` to allow configuring [GraphQL's authentication](https://www.apollographql.com/docs/react/networking/authentication/#cookie) when the apollo client (`GRAPHQL_SCHEMA_URL`) is external.

**Note: if the environment variables are not set, there aren't any behavior changes.**

### DIFF
--- a/packages/dashboard/nginx/default.conf.template
+++ b/packages/dashboard/nginx/default.conf.template
@@ -2,6 +2,6 @@ server {
 	listen ${PORT} default_server;
 	root /usr/share/nginx/html;
 	try_files $uri /index.html;
-	sub_filter '<?- SORRY_CYPRESS_ENVIRONMENT ?>' '{ GRAPHQL_CLIENT_ATTRIBUTES: "${GRAPHQL_CLIENT_ATTRIBUTES}", GRAPHQL_SCHEMA_URL: "${GRAPHQL_SCHEMA_URL}", CI_URL: "${CI_URL}" }';
+	sub_filter '<?- SORRY_CYPRESS_ENVIRONMENT ?>' '{ GRAPHQL_CLIENT_CREDENTIALS: "${GRAPHQL_CLIENT_CREDENTIALS}", GRAPHQL_SCHEMA_URL: "${GRAPHQL_SCHEMA_URL}", CI_URL: "${CI_URL}" }';
 	sub_filter_once on;
 }
--- a/packages/dashboard/nginx/default.conf.template
+++ b/packages/dashboard/nginx/default.conf.template
@@ -2,6 +2,6 @@ server {
 	listen ${PORT} default_server;
 	root /usr/share/nginx/html;
 	try_files $uri /index.html;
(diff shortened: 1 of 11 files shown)


agoldis (changes requested):
Hey @fsmaia thanks for that PR! It is a really good idea. 

It requires a bit more work to use environment variables on dashboard. It runs in browsers. See how `CI_URL` and `GRAPHQL_SCHEMA_URL` are implemented. Also note https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/dashboard/nginx/default.conf.template

packages/dashboard/nginx/default.conf.template
```diff
@@ -2,6 +2,6 @@ server {
 	listen ${PORT} default_server;
 	root /usr/share/nginx/html;
 	try_files $uri /index.html;
-	sub_filter '<?- SORRY_CYPRESS_ENVIRONMENT ?>' '{ GRAPHQL_SCHEMA_URL: "${GRAPHQL_SCHEMA_URL}", CI_URL: "${CI_URL}" }';
+	sub_filter '<?- SORRY_CYPRESS_ENVIRONMENT ?>' '{ GRAPHQL_CLIENT_ATTRIBUTES: "${GRAPHQL_CLIENT_ATTRIBUTES}", GRAPHQL_SCHEMA_URL: "${GRAPHQL_SCHEMA_URL}", CI_URL: "${CI_URL}" }';
```
agoldis, line 5: Now there's a different problem - if it isn't defined, the script fails :) Try to actually run the docker image

agoldis: I was trying to find a way to have 'default' values for ngingx template. 
The problem is that if variable isn't defined, it fails. That mean we have to defined env vars in all docker-compose and helm charts. Very inconvenient. Any ideas?

fsmaia: Yep :/
Any suggestions on how to proceed?

agoldis: May be just setting the defaults in the Dockerfile?

```
ENV PORT 80
ENV CI_URL https://host.docker.internal:5000
```

Do you know if docker-compose is able to override those after the image had been built?

tico24: Yes. It will. 

So will docker run and k8s

agoldis: @fsmaia would you mind defining the defaults then.
How soon do you want this merged? I wanted to actually test it tonight

fsmaia: Tomorrow would be fine to merge, I had some meetings but I'm free until the end of the day to work/help on this. Count on me :)

fsmaia: Now the defaults are in the Dockerfile, should we remove them from the docker-compose files when the value equals the default?

agoldis: Sorry for delayed response. Since you've already added the variables, let's keep those!

agoldis (approved):
Nicely done! 🎉🚀

PR #309, opened 2021-04-06
Add slack hook filters and advanced formatting

DeniDoman:
We have faced with 3 issues in a current slack flow:

1. We want to get notification about only failed builds, not all builds.
2. We have dozens of branches, we want to know only about failed builds from **develop**, **master** and **release-\*** branches.
3. We need more information in slack message (commit message), also need to fix confusion with skipped/pending (https://github.com/cypress-io/cypress/issues/3092)

So, I have added:

1. Result filter (ALL, ONLY_FAILED, ONLY_SUCCESSFUL). Required field, default value is ALL.  
2. Branch filter (with '*' and '?' wildcards support). Optional field, empty state means filter is off, all branches are tracked. 
3. New Slack message layout.

Edit hook before:
image

Edit hook after:
image

Slack message before:
image

Slack message after:
Passed:
image
Failed:
image

### DIFF
--- a/packages/api/src/schema/schema.graphql
+++ b/packages/api/src/schema/schema.graphql
@@ -55,7 +55,7 @@ type SpecStats {
 
 scalar GenericHookType
 scalar SlackHookType
-scalar SlackEventFilter
+scalar SlackResultFilter
 scalar GithubHookType
 scalar BitbucketHookType
 
@@ -74,13 +74,13 @@ type SlackHook {
   url: String!
   hookType: SlackHookType!
   hookEvents: [String!]!
-  slackEventFilter: SlackEventFilter!
+  slackResultFilter: SlackResultFilter!
   slackBranchFilter: [String]
 }
 
 input CreateSlackHookInput {
   projectId: ID!
-  slackEventFilter: SlackEventFilter = "ALL"
+  slackResultFilter: SlackResultFilter = "ALL"
 }
 
 input UpdateSlackHookInput {
@@ -88,7 +88,7 @@ input UpdateSlackHookInput {
   hookId: ID!
   url: String!
   hookEvents: [String!]!
-  slackEventFilter: SlackEventFilter!
+  slackResultFilter: SlackResultFilter!
   slackBranchFilter: [String]
 }
 
@@ -168,7 +168,7 @@ type Hook {
   bitbucketUsername: String
   bitbucketToken: String
   bitbucketBuildName: String
-  slackEventFilter: String
+  slackResultFilter: String
   slackBranchFilter: [String]
 }
 
@@ -189,7 +189,7 @@ input HookInput {
   bitbucketUsername: String
   bitbucketToken: String
   bitbucketBuildName: String
-  slackEventFilter: String
+  slackResultFilter: String
   slackBranchFilter: [String]
 }
 
--- a/packages/api/src/schema/schema.graphql
+++ b/packages/api/src/schema/schema.graphql
@@ -55,6 +55,7 @@ type SpecStats {
 
 scalar GenericHookType
 scalar SlackHookType
+scalar SlackEventFilter
 scalar GithubHookType
 scalar BitbucketHookType
 
@@ -73,10 +74,13 @@ type SlackHook {
   url: String!
   hookType: SlackHookType!
   hookEvents: [String!]!
+  slackEventFilter: SlackEventFilter!
+  slackBranchFilter: [String]
 }
 
 input CreateSlackHookInput {
   projectId: ID!
+  slackEventFilter: SlackEventFilter = "ALL"
 }
 
 input UpdateSlackHookInput {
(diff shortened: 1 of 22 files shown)


DeniDoman:
Please don't merge. Regex idea was not very good: branches with name like "agg23/Typescript4.2.3" doesn't work correctly in filter.

packages/director/src/lib/hooks/utils.ts
```diff
@@ -1,24 +1,80 @@
...
+    if (!hook.hookEvents || !hook.hookEvents.length) {
+      return true;
+    }
+
+    if (hook.hookEvents.includes(event)) {
+      return true;
+    }
   }
 
   return false;
```
agoldis, line 70: That becomes a bit bulky to read, would you mind please extracting to a separate function, smth like `shouldReportSlackHook`

packages/director/src/lib/hooks/reporters/controller.ts
```diff
@@ -54,6 +55,7 @@ export function reportToHook({
         reportToGenericWebHook({
           hook,
           runId: run.runId,
+          branch: run.meta.commit.branch,
```
agoldis, line 58: 👍🏻

packages/api/src/schema/schema.graphql
```diff
@@ -181,6 +189,8 @@ input HookInput {
   bitbucketUsername: String
   bitbucketToken: String
   bitbucketBuildName: String
+  slackEventFilter: String
+  slackBranchFilter: [String]
```
agoldis, line 193: That allows `[null]`, which is not something we want I guess :) 

How about `[String!]`?

packages/api/src/schema/schema.graphql
```diff
@@ -162,6 +168,8 @@ type Hook {
...
```
agoldis, line 172: That's very unfortunate duplication :( Would be nice to eliminate it eventually

agoldis (commented):
Very nice @DeniDoman! Thanks a lot for this PR!

 I have added a few minor comments.

packages/api/src/schema/schema.graphql
DeniDoman: I see that the whole Hook type doesn't contain required fields. I thought this some tricky plan, so I leave new fields not required too. 

```
type Hook {
  hookId: String
  url: String
  headers: String
  hookEvents: [String]
  hookType: String
  githubToken: String
  githubContext: String
  bitbucketUsername: String
  bitbucketToken: String
  bitbucketBuildName: String
  slackEventFilter: String
  slackBranchFilter: [String]
}
```

Should I add a required flag only to my fields or for all relevant fields? BTW branch filter can be empty or null, but slackEventFilter must have a value.

packages/director/src/lib/hooks/reporters/slack.ts
```diff
@@ -1,46 +1,67 @@
...
+    } *Skipped:* ${pending}\n\n\n` +
+    `${failures + skipped > 0 ? ':red_circle:' : ':white_circle:'} *Failed*: ${
+      failures + skipped
+    }`;
+
+  const commitDescription = `*Branch:*\n${branch}\n\n*Commit:*\n${
+    message.length > 100 ? `${message.substring(0, 100)}...` : message
+  }`;
 
   if (hookEvent === HookEvent.RUN_FINISH) {
```
DeniDoman, line 64: @agoldis could you please explain why we need this 5 seconds timeout? Can't get it.

DeniDoman:
```
  if (hookEvent === HookEvent.RUN_FINISH) {
    await new Promise((resolve) => {
      setTimeout(() => {
        resolve(null);
      }, 5000);
    });
  }
```

packages/api/src/schema/schema.graphql
DeniDoman: Also empty or null slackBranchFilter means all branches are tracked. So it's a valid state.

packages/director/src/lib/hooks/reporters/slack.ts
DeniDoman: I have removed it so far. Say if need to revert.

packages/api/src/schema/schema.graphql
DeniDoman: Agree, but seems as out of scope of the change.

packages/api/src/schema/schema.graphql
DeniDoman: I have left it as is.

packages/director/src/lib/hooks/utils.ts
DeniDoman: Totally agree, thank you. I have split it into 4 functions. Also I have covered all of them by unit tests 😎

packages/director/src/lib/hooks/reporters/slack.ts
agoldis: Doesn't seem like those are required.

agoldis (approved):

PR #320, opened 2021-04-21
Adds the ability to set an expireAfterSeconds (ttl) on documents

gareth4192:
Adds the ability to set an expireAfterSeconds (ttl) on documents in a MongoDB index by setting environment variables.

If the environment variables are not set there will be no changes in behaviour.

### DIFF
--- a/packages/api/src/config.ts
+++ b/packages/api/src/config.ts
@@ -8,7 +8,3 @@ export const MONGODB_USER = process.env.MONGODB_USER || undefined;
 export const MONGODB_PASSWORD = process.env.MONGODB_PASSWORD || undefined;
 export const MONGODB_AUTH_MECHANISM =
   process.env.MONGODB_AUTH_MECHANISM || undefined;
-
-export const RUNS_TTL_OPTIONS = { expireAfterSeconds: Number(process.env.RUNS_TTL) } || {};
-export const INSTANCES_TTL_OPTIONS = { expireAfterSeconds: Number(process.env.INSTANCES_TTL) } || {};
-export const PROJECTS_TTL_OPTIONS = { expireAfterSeconds: Number(process.env.PROJECTS_TTL) } || {};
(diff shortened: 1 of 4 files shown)


agoldis (commented):
Awesome! Thanks a lot, for that PR. That's a great and simple solution for data retention! 

I have a few comments though, please let me know what do you think:

- Please also update the documentation at https://github.com/sorry-cypress/gitbook. You can only add the config options for `director`, I am planning to remove indexes creation from the API anyways.
- There will be more collections soon - it would make sense to have a default TTL value for all collections. I imaging overriding TTL for individual collections would still be handy.
- Please scope the env variables name to `MONGO_XXX_SECONDS` for clarity

gareth4192:
Thank you for reviewing this, I have updated to docs, changed the env variables names, and removed the options from the API.

agoldis:
Thanks @gareth4192!

I did take a deeper look at the feature today. Here're a few thoughts:


- TTL settings only apply to indexes on ["_field that holds values of BSON date type_"](https://docs.mongodb.com/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time). The current implementation doesn't really expire documents. We can create an index on `createdAt` and convert it to a proper `Date` instead of `string`

- Converting the field to `Date` would require a workaround for legacy data. Otherwise it would be a breaking change.

- Applying the option `expireAfterSeconds` on an existing index fails - i.e. it would prevent seamless upgrade for existing users.

- Simply changing the configuration would not change the TTL value. We need to detect the change and "_modify ... the existing TTL index using the`collMod` command._".


Given those discoveries, it seems like there's more work needed to implement this feature.

gareth4192:
I see what you mean, thank you for taking the time to look into it deeper, i had misunderstood exactly how the expireAfterSeconds was working.
I’ll close this PR and take a look at the options, like you say above it looks it would be a breaking change to get this working so it may be that i look at another solution.
Thank you again for your time on this.

Issue #133, opened 2020-09-10
'Aggregation project operator not supported' error from API component (when using AWS DocumentDB)


joe-carpenter:
NB this deployment is currently configured using AWS DocumentDB (MongoDB 3.6 'api compatibility') - we have been running on-cluster Mongo up until now, but would like to make it a bit more stable and scalable as we are using `sorry-cypress` quite heavily.

All `sorry-cypress` components are latest versions.

When the frontend loads the main `All Runs` page, it sends a call to the API component with the following payload:

```
{"operationName":"getRunsFeed","variables":{"cursor":""},"query":"query getRunsFeed($cursor: String) {
  runFeed(cursor: $cursor) {
    cursor
    hasMore
    runs {
      runId
      createdAt
      meta {
        ciBuildId
        projectId
        commit {
          sha
          branch
          remoteOrigin
          message
          authorEmail
          authorName
          __typename
        }
        __typename
      }
      specs {
        spec
        instanceId
        claimed
        results {
          cypressConfig {
            video
            videoUploadOnPasses
            __typename
          }
          videoUrl
          tests {
            title
            state
            __typename
          }
          stats {
            tests
            pending
            passes
            failures
            skipped
            suites
            wallClockDuration
            wallClockStartedAt
            wallClockEndedAt
            __typename
          }
          __typename
        }
        __typename
      }
      __typename
    }
    __typename
  }
}
"}
```

which returns the following error:

```
{
    "errors": [
        {
            "message": "Aggregation project operator not supported: '$map'",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "runFeed"
            ],
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR",
                "exception": {
                    "ok": 0,
                    "code": 305,
                    "name": "MongoError",
                    "stacktrace": [
                        "MongoError: Aggregation project operator not supported: '$map'",
                        "    at Connection.<anonymous> (/app/node_modules/mongodb/lib/core/connection/pool.js:451:61)",
                        "    at Connection.emit (events.js:198:13)",
                        "    at processMessage (/app/node_modules/mongodb/lib/core/connection/connection.js:452:10)",
                        "    at TLSSocket.<anonymous> (/app/node_modules/mongodb/lib/core/connection/connection.js:621:15)",
                        "    at TLSSocket.emit (events.js:198:13)",
                        "    at addChunk (_stream_readable.js:288:12)",
[...]

PR #324, opened 2021-04-22
Remove mongo $map usages to simpler syntax

pbeckham:
$map is incompatible with the current version of Amazon DocumentDB, removing it allows sorry-cypress to be more easily deployed in AWS using the mongo "api-compatible" DocumentDB.

I have not used mongo heavily before, but from my testing this seems to give the same result.

Would potentially fix https://github.com/sorry-cypress/sorry-cypress/issues/133

### DIFF
--- a/packages/api/src/datasources/runs.ts
+++ b/packages/api/src/datasources/runs.ts
@@ -58,13 +58,7 @@ const projectAggregation = {
     specs: 1,
     createdAt: 1,
     completion: 1,
-    specsFull: {
-      $map: {
-        input: '$specs',
-        as: 'spec',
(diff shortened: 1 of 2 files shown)


agoldis:
@pbeckham thanks for that PR! Will review shortly

agoldis:
Awesome @pbeckham! Much simpler indeed! If you see other opportunities to simplify mongo queries - please do! 

As a side note, there's a chance future changes would not be 100% compatible with DocumentDB. "officially" supporting two implementations would create more maintenance  burden - I am trying to avoid that :)

Do you have any idea - how can we automate integration with DocumentDB? Also, moving this Q to discussions...

PR #325, opened 2021-04-23
feature: support monorepos for BitBucket hooks

Tom-Hudson:
Currently the commit sha is used as the key in the request to add/update the BitBucket status. The key is the externally generated unique reference used to identify the status via the BitBucket API.

Using the commit sha means that if you have a monorepo with multiple Sorry Cypress projects setup, the latest hook will override the status. For example, we have 4 projects in the same monorepo which all have BitBucket hooks setup. I have named them uniquely but because they all use the commit sha to add/update the status, they override each other. This is frustrating because it means you do not get a status for each of the projects, so if one fails it's failed status won't be shown if the latest one passes. 

This change uses the hookId (a UUID) as the key instead. This ensures the keys will be unique to each hook and will enable all 4 projects to add/update the correct status for the same commit (the commit is the entity being updated via the BitBucket API).

### DIFF
--- a/packages/director/src/lib/hooks/__tests__/fixtures/bitbucketReportStatusRequest.json
+++ b/packages/director/src/lib/hooks/__tests__/fixtures/bitbucketReportStatusRequest.json
@@ -9,7 +9,7 @@
     "Accept": "application/json"
   },
   "data": {
-    "key": "testCommitSha",
+    "key": "testHook",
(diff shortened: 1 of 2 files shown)


agoldis:
Oh, I think `hookId` was in the original implementation, but I misunderstood it and replaced with `sha` - my bad, sorry for that!

PR #343, opened 2021-05-09
fix: correctly extract ciBuildId from Gitlab CI

boxofcrates:
CI_BUILD_ID is no longer set by Gitlab CI, per fsharp/FAKE#2290.

Not referencing an issue as there is none filed, but I can file/reference if needed.

I have confirmed CI_PIPELINE_ID is a viable env var by running Cypress with explicit `--ci-build-id $CI_PIPELINE_ID` (which works) and `--ci-build-id $CI_BUILD_ID` (which does not).


agoldis:
Awesome @boxofcrates - thanks for tackling that!

PR #399, opened 2021-07-21
feature: allow reset of instance to facilitate retesting

Aeolun:
Basically, my tests sometimes fail due to flakiness, and I need some way to reset only a specific instance, instead of having to run the whole suite again.

This allows sorry cypress to run just the failed tests again by making it as if they haven't been picked up by any executor yet.

Happy to hear any comments you have.

### DIFF
--- a/packages/dashboard/src/run/runDetails/resetInstance/resetInstanceButton.tsx
+++ b/packages/dashboard/src/run/runDetails/resetInstance/resetInstanceButton.tsx
@@ -1,0 +1,131 @@
+import {
+  GetRunDocument,
+  useResetInstanceMutation,
+} from '@src/generated/graphql';
+import { useAsync } from '@src/hooks/';
+import {
+  Alert,
+  Button,
+  Heading,
+  HFlow,
+  Icon,
+  Modal,
+  ModalBody,
+  ModalFooter,
+  Text,
+} from 'bold-ui';
+import React, { useCallback, useEffect, useState } from 'react';
+import { useRouteMatch } from 'react-router-dom';
+
+export const ResetInstanceButton = ({
+  instanceId,
+  runId,
+}: {
+  instanceId: string;
+  runId: string;
+}) => {
+  const {
+    params: { projectId },
+  } = useRouteMatch<{ projectId: string }>();
+
+  const [resetInstance] = useResetInstanceMutation({
+    variables: {
+      instanceId: instanceId,
+    },
+    refetchQueries: [
+      {
+        query: GetRunDocument,
+        variables: {
+          runId: runId,
+        },
+      },
+    ],
+  });
+
+  const [startDeleteInstance, deleting, deleteResult, deleteError] = useAsync(
+    resetInstance
+  );
+  const [shouldShowModal, setShowModal] = useState(false);
+
+  useEffect(() => {
+    if (!deleteResult) {
+      return;
+    }
+    setShowModal(false);
+  }, [deleteResult]);
+
+  const deleteAction = useCallback(
+    (e) => {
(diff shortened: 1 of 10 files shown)


packages/dashboard/src/run/runDetails/resetInstance/resetInstanceButton.tsx
```diff
@@ -0,0 +1,131 @@
...
+      return;
+    }
+    setShowModal(false);
+  }, [deleteResult]);
+
+  const deleteAction = useCallback(
+    (e) => {
+      e.stopPropagation();
+      e.preventDefault();
+      console.log('deleting here');
```
agoldis, line 61:
```suggestion
```

packages/dashboard/src/run/runDetails/resetInstance/resetInstanceButton.tsx
```diff
@@ -0,0 +1,131 @@
...
+              icon="exclamationTriangleFilled"
+              style={{ marginRight: '0.5rem' }}
+              size={3}
+              fill="danger"
+            />
+            <div>
+              <Heading level={1}>Delete instance {instanceId}?</Heading>
+              <Heading level={5}>
+                This will remove result information for this instance. You will
+                need to run the tests again to re-record data for this instance.
```
agoldis, line 86:
```suggestion
                This will remove result information for this instance. You will
                need to run the tests with the case ci-build-id again to re-record data for this instance.
```

agoldis (commented):
Awesome job @Aeolun! Very elegant solution!

agoldis:
@all-contributors please add @Aeolun for code

flipace:
@Aeolun @agoldis We're using CircleCI and currently pass the CIRCLE_BUILD_NUM as build id. When rerunning via CCI (eg on test flake), the build number is going to change so it would report this to a new job id in sorry-cypress.

We could use the commit SHA instead as the build id, but then we wouldn't be able to link from the build in sorry-cypress to the circleci test run since we need the build num and project id for that 🤔 

Am I missing some additional configuration option for the URL or is, in our case, this rerunning of a specific test not possible? 🤔

Aeolun:
Note: Writing from phone, I’ll have a look at the actual code later, if agoldis who probably knows better doesn’t reply first ;)

@flipace I think this is mostly for the situation where the build doesn’t get a new ID. If the build gets a new ID, then doesn’t that means that you already restarted the whole job?

If you want to associate the new run with the earlier build then I think you need to use a build ID that doesn’t change between runs.

I agree that’s inconvenient in this case, but I think it may be better to build a feature to associate multiple build jobs urls with a given build ID in sorry-cypress?

agoldis:
@flipace 

If you're looking to get a link from sorry-cypress run to CI job, I think it can be built from the metadata we have for each run. It's just not exposed in the UI. 

Take a look at `run.meta.ci.params`, here's an example for Azure:
```
"ci": {
            "params": {
                "buildBuildid": "4554",
                "buildBuildnumber": "20210721.5",
                "buildContainerid": "15297276",
                "buildRepositoryUri": "https://github.com/moneycorp/branchordering"
            },
            "provider": "azure"
        }
```

That mean you can use any other `ci-build-id` instead of `CIRCLE_BUILD_NUM`. May be something that doesn't change between reruns.

I have opened https://github.com/sorry-cypress/sorry-cypress/issues/414 to track it.

@Aeolun 

>  I think it may be better to build a feature to associate multiple build jobs urls with a given build ID in sorry-cypress?

That's quite tricky, I remember some requests in past - I am afraid of taking that direction because I don't know how to solve it without a significant rewrite.

PR #422, opened 2021-08-18
feat: allow linking directly to a project/build-id combination

Aeolun:
Small PR to do a thing that we're now constantly doing manually. I haven't found a way to link directly to a project/build-id combination, so I've added one.

I'm not really proud of the implementation, but I couldn't find a better way of implementing the redirect logic using react router.


agoldis:
Thanks for this PR @Aeolun it looks good, I will merge it and release with some other changes I've been working on.

PR #427, opened 2021-08-26
v2.0

agoldis:
v2.0

### Breaking changes
- Deprecated support for cypress agents lt `6.7.0`
  - Supporting the legacy versions of cypress with all the code was cumbersome. Trying to use SC with older cypress versions would return an error when creating new runs. Closes #412.

- The internal representation of runs has changed. **Runs created prior to v2.0 might be displayed partially or not displayed at all.**
  - added a `progress` field on `run` with the instances and tests progress state. We use this field to report run's progress in hooks / dashboard instead of invoking complex MongoDB queries. This should resolve #417 because we won't use MongoDB aggregations that create gt 16MB documents.
  - `runs.specs` will have a short version of "results" - that would allow more efficient data fetching for showing runs feeds and individual runs.
 
### Other changes

- fix: Delete run timeout when deleting run. Closes #409.
- fix: Correctly report failed tests w/o counting retires. Closes #384 
- fix: In-memory director crashes when test fails with an exception. Closes #425 
- fix: Stop showing duration running for completed runs / tests. Closes #377
- feat: Add retries to Slack integration, show retries count everywhere and use "Flaky" badge if spec / test was retried. Closes #378 
- feat: Configure default page items # on runs feed via `PAGE_ITEMS_LIMIT` env variable for API service
- infra: remove redis dependency in docker-compose files, updated docs accordingly
- infra: properly set up typescript for monorepo, resolved dozens of TS errors and warnings
- misc: completely removed lookup aggregations from mongoDB queries. Sorry cypress is much DocumentDB friendly now!
- misc: added material-UI for gradual transition. See #401

PR #433, opened 2021-09-07
Delete runs in daterange in batches

Roemer:
This PR changes the behavior of the deletion of runs in a specified date range.
Previously, it used a very slow `aggregate` to find the `instances` for the `runs` that should be deleted and tried to delete them all at once. This timeouted at various ends in larger setups.
With this change, it will now process the deletion in batches:
1. Get a batch of `runIds`
2. Delete all associated `instances` for those `runIds`
3. Delete the `runs` for those `runIds`
4. Repeat until no more `runs` are found for the defined date-range
5. 
To make it even faster, a new index for `runId` is added to the `instance` collection.

### DIFF
--- a/packages/api/src/datasources/instances.ts
+++ b/packages/api/src/datasources/instances.ts
@@ -85,29 +85,4 @@ export class InstancesAPI extends DataSource {
       instanceId: result.result.ok === 1 ? instanceId : undefined,
     };
   }
-
-  async deleteInstancesInDateRange(startDate: Date, endDate: Date) {
-    if (startDate > endDate) {
-      return {
-        success: false,
-        message: `startDate: ${startDate.toISOString()} should be less than endDate: ${endDate.toISOString()}`,
-        runIds: [],
-      };
-    }
-    const response = await Collection.instance()
-      .aggregate([
-        {
-          $match: {
-            'run.createdAt': {
-              $lte: endDate.toISOString(),
-              $gte: startDate.toISOString(),
-            },
-          },
-        },
-      ])
-      .toArray();
-
-    const runIds = response.map((x) => x.runId) as string[];
-    return await this.deleteInstancesByRunIds(runIds);
-  }
 }
--- a/packages/api/src/datasources/runs.ts
+++ b/packages/api/src/datasources/runs.ts
@@ -124,7 +124,16 @@ export class RunsAPI extends DataSource {
     };
   }
 
-  async deleteRunsInDateRange(startDate: Date, endDate: Date) {
+  async deleteRunsInDateRange(
+    startDate: Date,
+    endDate: Date,
+    limit: number = 0
+  ) {
+    const getResult = await this.getRunsInDateRange(startDate, endDate, limit);
+    return this.deleteRunsByIds(getResult.runIds);
+  }
+
+  async getRunsInDateRange(startDate: Date, endDate: Date, limit: number = 0) {
     if (startDate > endDate) {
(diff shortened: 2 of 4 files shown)


packages/api/src/datasources/runs.ts
```diff
@@ -136,23 +136,37 @@ export class RunsAPI extends DataSource {
...
 
-  async deleteRunsInDateRange(startDate: Date, endDate: Date) {
+  async deleteRunsInDateRange(
+    startDate: Date,
+    endDate: Date,
+    limit: number = 0
+  ) {
+    const getResult = await this.getRunsInDateRange(startDate, endDate, limit);
+    return this.deleteRunsByIds(getResult.runIds);
+  }
```
agoldis, line 146: This one is not used, please delete

agoldis (approved):
Great job @Roemer! Thanks a lot for fixing that code I am happy to see that people use Sorry Cypress to run as many test that we are facing performance issues 😬

Just a few minor comments:

- batched iteration on results is great and simple! I think the same result could be achieved with native mongoDB [cursor](https://docs.mongodb.com/drivers/node/fundamentals/crud/read-operations/cursor/#asynchronous-iteration), but your implementation is totally find!
- if we are really looking to optimize it, we can limit the fields we get from `getRunsInDateRange` by explicitly [fetching **only**](https://docs.mongodb.com/drivers/node/fundamentals/crud/read-operations/project/#specify-which-fields-to-return) `runId`s.   If I am not mistaken, mongoDB won't even read data from the disk and send the results back straight from the `runId` index on runs collection.

packages/api/src/datasources/instances.ts
```diff
@@ -89,32 +89,6 @@ export class InstancesAPI extends DataSource {
...
-              $gte: startDate.toISOString(),
-            },
-          },
-        },
-      ])
-      .toArray()) as InstanceWithRuns[];
-
-    const runIds = response.map((x) => x.run[0].runId) as string[];
-    return await this.deleteInstancesByRunIds(runIds);
-  }
```
agoldis, line 117: 👍🏻

packages/api/src/resolvers/index.ts
```diff
@@ -160,15 +160,36 @@ export const resolvers = {
...
+          startDate,
+          endDate,
+          batchSize
+        );
+        if (getRunsResponse.runIds.length == 0) {
+          // No more runs in the range, break out of the loop
+          break;
+        }
+        // Delete the runs
+        const runsDeleteResponse = await resolvers.Mutation.deleteRuns(
```
agoldis, line 177: Nice reuse of mutation 👍🏻

agoldis:
P.S. I've merged #427 into master, don't be scared

Roemer:
I rebased it back to master instead of merged and removed the unneeded method and added a projection when getting the runs.
Good point about the cursor. I had a look at it and it would not really fit well into the current architecture. We would need to add the cursor to datasources/run.ts and actually return it so that we can handle the deletion of the instances, runs and runtimeouts in the mutation itself. But then we would have a datasource related object (the cursor) in the mutation itself which seems wrong. Or we would pass a function to the run datasource which would handle that all, but that also seems not so nice. So I think for a first version, we can go with this.
Testing is still pending on my side. I will probably need to do that tomorrow.

Roemer:
I now tested it and I could delete 2 months worth of data in a few seconds ;) So from my side, this PR can be merged.

Issue #401, opened 2021-07-22
🙏🏻 Help Needed: boldui -> Material-UI


YeonV:
As part of Sorry Cypress v2.0, we are planning to migrate the dashboard to material UI as requested.

That's quite a big effort and I need your help. I believe that gradually we can migrate the whole project to Material UI. If you are interested in helping with migration, please comment below and start working on the corresponding screen.
 
@YeonV @conantteo @ImanMahmoudinasab @fsmaia 

## Development guide

- Add a comment below your GH nickname and what component would you like to work on
- Please fork off https://github.com/sorry-cypress/sorry-cypress/tree/feat/v2.0
- See [Development Guide](https://docs.sorry-cypress.dev/development/development-guide) to bootstrap the project
- Wrap the modified component in `WithMaterial` HOC. [Example](https://github.com/sorry-cypress/sorry-cypress/blob/feat/v2.0/packages/dashboard/src/project/projectsView.tsx)
- Open a PR to https://github.com/sorry-cypress/sorry-cypress/tree/feat/v2.0

## List of screens / components + volume of work
- [x] ~Update to material ui v5 (@agoldis)~
- [x] ~Layout (sm) (@ImanMahmoudinasab)~
- [x] ~Navbar (sm) (@ImanMahmoudinasab)~
- [x] ~Projects List (sm) (@ImanMahmoudinasab)~
- [ ] Projects Edit (md) (@karimould)
- [ ] Edit / Create Hooks (lg) (@karimould)
- [x] Runs Feed Screen (sm) (@ImanMahmoudinasab) 
- [x] Run Summary Component (md) (@ImanMahmoudinasab)
- [x] Run Details Screen (lg) (@ImanMahmoudinasab)
- [x] Spec Details Screen (md) (@ImanMahmoudinasab)
- [x] Test Details Screen (md) (@ImanMahmoudinasab)
- [x] Run / Spec / Test status badges (sm) (@ImanMahmoudinasab)
- [x] Run Progress Component (sm) (@ImanMahmoudinasab)
- [ ] Landing Page (sm) (@ImanMahmoudinasab)

p.s.
Sorry @YeonV I am piggybacking this issue.

agoldis:
Thanks for opening the issue, I am totally down for moving to a better framework, please, Material UI or any other popular. It'd required some significant amount of refactoring though. Any help appreciated

ImanMahmoudinasab:
here are some new designs, feedbacks appreciated:

[...]

PR #459, opened 2021-09-26
Migrate projects list to MUI

ImanMahmoudinasab:
- [x] #401 

image

image

image

image

image


packages/dashboard/src/project/projectList.tsx
```diff
@@ -15,7 +17,7 @@ const ProjectsList = ({ search }: ProjectsListProps) => {
   const filters: Filters[] = search
     ? [
         {
-          value: null,
+          value: (undefined as unknown) as null,
```
ImanMahmoudinasab, line 20: This fixes an existing bug which cause searching projects not returning any result.

packages/dashboard/src/run/runDetails/resetInstance/resetInstanceButton.tsx
```diff
@@ -24,10 +23,6 @@ export const ResetInstanceButton = ({
   instanceId: string;
   runId: string;
 }) => {
-  const {
-    params: { projectId },
-  } = useRouteMatch<{ projectId: string }>();
-
```
ImanMahmoudinasab, line 30: Removed unused codes to get rid of eslint errors.

packages/dashboard/src/project/projectList.tsx
ImanMahmoudinasab: It is not the best way to solve the issue but the easiest. We should make this prop optional for `filters` as may someone need to pass `like` prop not the `value` prop.

agoldis (approved):
@ImanMahmoudinasab Loving it a lot! ❤️🎉

PR #474, opened 2021-11-04
Upgrade instance and test view to MUI

ImanMahmoudinasab:
New features:

- [x] Discuss feature #401 

image

image

image

image

image

image


agoldis (commented):
Looks great @ImanMahmoudinasab, thanks again for your amazing work! Here are a few comments from my side:

Screen Shot 2021-11-05 at 10 50 51 PM
Screen Shot 2021-11-05 at 10 49 52 PM
Screen Shot 2021-11-05 at 10 41 53 PM



The following suggestion are not part of this PR though. Please tell me what you think... We can tackle those issues in a follow up PR.


Screen Shot 2021-11-05 at 10 47 33 PM
Screen Shot 2021-11-05 at 10 35 35 PM
Screen Shot 2021-11-05 at 10 35 16 PM

ImanMahmoudinasab:
**Failed tests messages:**

> monospace?

No they are not monospace, but `white-space: pre-wrap`. 

With `white-space: pre-wrap`:
image

Without `white-space: pre-wrap`: 
image

ImanMahmoudinasab:
> When shrink to md or smaller test list collapses

the test list will fill all the space available, and the test result will be fell bellow the test lis. This was due to inadequate space for showing both tests list and test result, specially when sidebar is open.

When I decide to make them full-width on <= md screens I hadn't add the collapse button for tests list. What I will do is making the test list full width in smaller screen sizes.

ImanMahmoudinasab:
>  image

In run list, I think there shouldn't be any link inside the paper and clicking everywhere of the card should navigate user to the details of the run. This will make it easy to click on the card without thinking in which spot of card you shouldn't click because it would not go to run detail view. 

This was my reason for this design, if you still think it is better to keep them as links, I can change the behaviour.

>  Should it be a url? 
>
> image

Here I thought most of the time user would know the origin address (at least for me is like this ;) ) so instead of showing a [long] url which will get extra attention is not what is desired by users. If they need to know the origin they can , hover over it or click on it.(we can show it on the tooltip).

Again I can change it if you think we should do that.

ImanMahmoudinasab:
I changed the breakpoints to full width the test list and test result on small screens:

image

on md size screen size user can click on collapse button:
2021-11-06 at 10 04 AM

result:
image

ImanMahmoudinasab:
image

image

image

image

image

ImanMahmoudinasab:
@agoldis Comments applied.

PR #490, opened 2021-11-14
Update dashboard packages

ImanMahmoudinasab:
Some of updated packages:
-  "@mui/icons-material": "^5.1.0"
- "react-router": "^6.0.2"
- "webpack": "^5.64.0"
- "webpack-dev-middleware": "^5.2.1"
- "webpack-dev-server": "^4.5.0"


fsmaia:
This breaks projects with slashes in the name:
https://reactrouter.com/docs/en/v6/upgrading/v5#note-on-route-path-patterns

Could we revert to react-router v5?

cc @agoldis

PR #495, opened 2021-11-17
updated the CloudFormation Stack with proper health checks

pitthecat:
Hi there, I updated the health checks in the CloudFormation stack as they were a bit cumbersome. 

- The Apollo webserver for GraphQL offers `/.well-known/apollo/server-health` as [health path](https://www.apollographql.com/docs/apollo-server/monitoring/health-checks/) -> you get HTTP 200 as return value which is way more intuitive matcher than 400 :) 
- I also updated the Director health check to `/ping` which allows the usage of HTTP 200 as matcher

I increased `HealthCheckIntervalSeconds` to 15 seconds. This is more than enough and you have less log entries with simple health check GET requests in AWS Cloudwatch. I deployed the improved health checks on my cluster today and they work fine.

### DIFF
--- a/cloudformation/sorry-cypress.yml
+++ b/cloudformation/sorry-cypress.yml
@@ -208,8 +208,8 @@ Resources:
   TargetGroupDirector:
     Type: AWS::ElasticLoadBalancingV2::TargetGroup
     Properties:
-      HealthCheckIntervalSeconds: 6
-      HealthCheckPath: /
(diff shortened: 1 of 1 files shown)


agoldis:
@all-contributors please add @pitthecat pitthecat for infrastructure

agoldis:
Awesome  @pitthecat, thanks you very much!

PR #509, opened 2021-12-13
Revert react-router to v5

fsmaia:
For bug fixes:

- [x] Add a reference to the original issue
- [ ] Add a test and/or some kind of instructions to verify the fix

Fixes a regression introduced by https://github.com/sorry-cypress/sorry-cypress/pull/490.

Note: I'm unable to run the project locally, I need help testing.

---

For new features:

- [ ] It's better to first discuss features proposal in [Discussions](https://github.com/sorry-cypress/sorry-cypress/discussions) or [Slack](https://sorry-cypress.slack.com/join/shared_invite/zt-eis1h6jl-tJELaD7q9UGEhMP8WHJOaw#/)
- [ ] Describe the use-case in details
- [ ] Ensure the solution is backwards-compatible
- [ ] Test it. Submit screenshots / outputs / tests together with the PR.

--- Side note ---

Your contribution is very valuable, we **really appreciate your time and effort**!

Please note:

- Adding a new code and / or feature complicates the maintenance
- Most people disappear after contributing a feature, leaving the support to core maintainers

Given that, please don't be offended if your PR is rejected or significantly modified.

### DIFF
--- a/packages/dashboard/package.json
+++ b/packages/dashboard/package.json
@@ -51,8 +51,8 @@
     "react-dom": "^17.0.2",
     "react-hook-form": "^6.15.8",
     "react-player": "^2.9.0",
-    "react-router": "^6.0.2",
-    "react-router-dom": "^6.0.2",
+    "react-router": "^5.1.2",
+    "react-router-dom": "^5.1.2",
     "react-transition-group": "^4.4.2",
     "string-hash": "^1.1.3"
   },
--- a/packages/dashboard/src/dashboard/dashboardView.tsx
+++ b/packages/dashboard/src/dashboard/dashboardView.tsx
@@ -4,12 +4,12 @@ import { Toolbar } from '@sorry-cypress/dashboard/components';
 import { setNav } from '@sorry-cypress/dashboard/lib/navigation';
 import ProjectsList from '@sorry-cypress/dashboard/project/projectList';
 import React, { useLayoutEffect, useState } from 'react';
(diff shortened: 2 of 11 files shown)


fsmaia:
As I can't run locally, I can't figure out if `useParams` will work fine. Could you help me to test this?

Just start the project locally, create a project with slug, and try to access it.

fsmaia:
Update: ran it locally (it was necessary to set Node version to v14), and it worked fine:
image

ImanMahmoudinasab:
Hi Fernando!
My thoughts on the issue:
If we are going to use a value in the url it shouldn't include characters that can make issue if we put the value in the url. One solution could be not allowing the special characters, so we will be sure the name is url-friendly. 
The other solution, which you mentioned on my PR, is encoding the value so users can use special characters if they have to. Yes it makes the url to not look so nice but it keeps url format valid. 
Overall, I believe by looking at url and seeing `/` in the url user should have only one perception.

I don't think it is good idea to reveart the router to v5 to just allow having `/` in url as part of a variable and see it `/` instead of `%3F`.

@agoldis What 's your idea?

agoldis:
I would go with encoded characters i.e. `%3F`

ImanMahmoudinasab:
I will create a pr and add the encoding and decoding logic to wherever we need it.
@fsmaia please feel free to create a pr if you have the chance to fix the issue before I start.

fsmaia:
Ok, we may close this one then :)
Big thanks for the discussion!

PR #523, opened 2022-01-17
Dockerfile efficiencies

tico24:
Shrink nodejs containers by using multistage builds more effectively:

* api: 306MB > 147MB
* director: 445MB > 184MB

Use [tini](https://github.com/krallin/tini) to ensure that nodejs processes respond to things like `SIGINT`.

Use `node` user instead of `root` user


tico24:
For what it's worth, I have built these containers locally and we are using them in production (testing in production is cool, right?). They seem to be running fine for us.

agoldis (approved):

PR #554, opened 2022-03-22
Adding proxy support for Webhooks

rene1039:
Use Case:
If a Sorry-Cypress instance is behind a proxy (and needs to be run behind a proxy), theres no way to send webhooks without using proxy. This solution is just an example for Slack Webhook, which worked in our environment. If interested, I can include the solution to other available webhooks.

### DIFF
--- a/packages/common/src/hook/types.ts
+++ b/packages/common/src/hook/types.ts
@@ -27,6 +27,9 @@ export enum ResultFilter {
 type BaseHook = {
   hookId: string;
   url: string;
+  proxyUrl;
+  proxyUsername;
+  proxyPassword;
 };
 
 export type SlackHook = BaseHook & {
--- a/packages/common/src/hook/types.ts
+++ b/packages/common/src/hook/types.ts
@@ -27,9 +27,6 @@ export enum ResultFilter {
 type BaseHook = {
   hookId: string;
   url: string;
-  proxyUrl;
-  proxyUsername;
-  proxyPassword;
 };
 
 export type SlackHook = BaseHook & {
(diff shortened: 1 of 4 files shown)


packages/common/src/hook/types.ts
```diff
@@ -27,6 +27,9 @@ export enum ResultFilter {
 type BaseHook = {
   hookId: string;
   url: string;
+  proxyUrl;
+  proxyUsername;
+  proxyPassword;
```
agoldis, line 32: It doesn't have any typing, you'd better remove alltogether

packages/director/src/lib/hooks/reporters/slack.ts
```diff
@@ -93,6 +94,20 @@ export async function reportToSlack(
         length: 100,
       }
     )}`;
+  
+  hook.proxyUrl = process.env.PROXY_URL
+  hook.proxyUsername = process.env.PROXY_USERNAME
+  hook.proxyPassword = process.env.PROXY_PASSWORD
+
+  if (hook.proxyUrl && hook.proxyUsername && hook.proxyPassword) {
```
agoldis, line 102: Please use `process.env` directly - no need to assign to `hook` properties

rene1039:
@agoldis Hi Andrew, I adjusted your inline comments as well as creating a MR for Sorry-Cypress Docu: https://github.com/sorry-cypress/gitbook/pull/11

How shall we extract the axios client? Do you have any solution or preferred way for this?

agoldis:
> How shall we extract the axios client?  Do you have any solution or preferred way for this?

like this https://axios-http.com/docs/instance

rene1039:
@agoldis I put the axios.create to the controller, so it will be checked when a new hook arrives.

PR #589, opened 2022-06-21
Add mongodb health checks

Hackatosh:
**New feature**

### Use case :
In certain case, the connexion to the MongoDB is lost and the director / GraphQL API must be rebooted in order for the connexion to work properly again.
This PR aims to implement a health check which allows to check if the connexion with the database is healthy. This health check makes it possible to automate the restart of the director / GraphQL API, through a Kubernetes liveness probe as an example.

### Implementation :
A route `/health-check-db` is added on both the director and the GraphQL API. The health check use the ping command (https://www.mongodb.com/docs/manual/reference/command/ping/) provided by MongoDB.
The implementation is very simple for the director as we just add an another route on the express application.
For the GraphQL API in order to add a route, we need to use apollo-server-express instead of apollo-server (NB : apollos-server in a wrapper made around apollo-server-express in order to make it easier to bootstrap an Apollo Server). 
In order to have full backward-compability for the GraphQL API, the code used to initialize the http server in apollo-server  has been copy pasted on sorry-cypress (original code can be found here, in the `listen` method : https://github.com/apollographql/apollo-server/blob/release-2.25.0/packages/apollo-server/src/index.ts) and changed where needed in order to add the new route.

### Test :
It is very easy to test the probe in using the provided docker-compose : after launching the docker-compose, the route should respond 200 for both the Director and the GraphQL API. After that, just kill the MongoDB container and both health check route should respond a 503.

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -55,13 +55,15 @@
     "@sorry-cypress/common": "1.0.0",
     "@sorry-cypress/logger": "1.0.0",
     "@sorry-cypress/mongo": "1.0.0",
-    "apollo-server": "^2.25.3",
+    "apollo-server-express": "^2.25.3",
     "dotenv": "^8.1.0",
+    "express": "^4.18.1",
     "graphql": "^15.3.0",
     "graphql-iso-date": "^3.6.1",
     "lodash": "^4.17.21",
     "plur": "^4.0.0",
     "source-map-support": "^0.5.20",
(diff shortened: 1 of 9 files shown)


agoldis:
@all-contributors please add @Hackatosh  for code

agoldis (approved):

agoldis:
Thanks a lot for your contribution @Hackatosh

Issue #74, opened 2020-05-15
Azure support


elkorba:
Hi,
are there any plans to support one of the Azure Storage services for screenshot driver?
If yes then when we can expect it?

Regards

agoldis:
Currently there're no plans for Azure - if you have expertise / time contributions are appreciated!

weyert:
The easiest way to accomplish this without code changes would be to put Minio in between as the S3 compatible API which then talks to Azure Storage: https://docs.min.io/docs/minio-gateway-for-azure.html

An alternative would be to write an abstraction so you can write ones for Azure Blob Storage, Google Blob Storage and one for the S3 compatible ones

agoldis:
FYI minio driver is now available https://github.com/agoldis/sorry-cypress/releases/tag/v0.6.0

onimougwo:
Still no plan for Azure support ? My company uses Azure Devops for CI, does that mean I can't use sorry-cypress ?

agoldis:
@onimougwo of course you can. It's just a docker images that you can deploy on any cloud provider. Unfortunately I don't  use Azure personally so I didn't write documentation. May be @weyert can help?

DanielleCogs:
@weyert any new ideas on Azure Storage use? Started going through the setup for MinIO Azure Gateway but looks like they will stop supporting it soon [(related MinIO issue)](https://github.com/minio/minio/issues/14331).

weyert:
@DanielleCogs Sorry, I stopped using Cypress

roldengarm:
What are the implications when running this on Azure? Setting up the Docker image is not a problem, but obviously I'd like to see the screenshots as well. With MinIO gone, is there an alternative?

DanielleCogs:
Yes, setting up the Docker image is fine. I was trying to use the AWS Cloud Formation setup as a template to use the equivalent Azure resources, but without S3 compatibility and MinIO being deprecated, we weren't able to deploy.

anilpujaraofficial:
how to give custome port for sorry cypress dashboard?

PR #595, opened 2022-06-28
Add Azure Blob Storage support

Hackatosh:
### Use case :

This PR aims to add support for Azure Blob Storage for screenshots / video uploads. Link to issue : https://github.com/sorry-cypress/sorry-cypress/issues/74

### Implementation :

I implemented the support for Azure Blob Storage using the ScreenshotsDriver interface. It was very straight-forward.

I also added a `docker-compose.azure-blob-storage.yml` which allows to start a full environment with a Azure Blob Storage container, similar to what is done is `docker-compose.minio.yml` and `docker-compose.cos.yml`.

Also, I remarked that there is a lot of duplicated code in the files related to screenshots drivers. I think it would be a good idea to refactor that in an other PR. What do you think ?

### Test :

In order to test this feature easily, I created another PR which add a route `generate-upload-urls-blob-storage` to the director and a Azure Blob Storage container to the docker-compose.build.yml environment : https://github.com/Hackatosh/sorry-cypress/pull/3

Using this branch, you can test the feature by following those steps :

- Rebuild your docker images declared in docker-compose.build.yml :
`docker-compose -f docker-compose.build.yml up --build`

- Call the route `generate-upload-urls-blob-storage` in order to get the uploadUrl and the readUrl. The easiest way to do that is to go to this adress `http://localhost:1234/generate-upload-urls-blob-storage?key=test` in your favorite browser

- Use the following curl in order to upload a test text file :
`curl --request PUT '<the-upload-url>' --header 'x-ms-blob-type: BlockBlob' --header 'Content-Type: text/plain' --data-binary '<path-to-a-text-file>'`

- Go to the read URL you have obtained earlier

You can also test with a real Azure Blob Storage container created on Azure. You just have to replace the `AZURE_CONNEXION_STRING` environment variable in the director container with the connexion string to you Azure Blob Storage Container on Azure.

### DIFF
--- a/docker-compose.azure-blob-storage.yml
+++ b/docker-compose.azure-blob-storage.yml
@@ -1,0 +1,77 @@
+version: '3.6'
+# https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio
+services:
+  mongo:
+    image: mongo:4.2
+    volumes:
+      - ./data/data-mongo-cypress:/data/db
+    ports:
+      - 27017:27017
+
+  director:
+    image: agoldis/sorry-cypress-director:latest
+    hostname: director
+    environment:
+      DASHBOARD_URL: http://localhost:8080
+      EXECUTION_DRIVER: '../execution/mongo/driver'
+      MONGODB_URI: 'mongodb://mongo:27017'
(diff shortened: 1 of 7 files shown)


agoldis:
@Hackatosh thanks a lot for that PR! Will review asap!

PR #608, opened 2022-07-24
Enhance sidebar layout

### LINKED ISSUE #581: Enhance sidebar layout
- [x] I have searched issues for similar feature requests

## What is this feature about (1 sentence)

The sidebar items and fonts are way too big to be on a side panel, and when the project has more than a few characters, it always produces a horizontal scroll.

## Why is it needed? What is the value? For whom do we build it?

The visual experience on the sidebar is not good and we have to use the horizontal scroll to find the projects.

## High-level feature overview

- Reduce the left padding on the project items (probably to 16px instead of 32px)
- Reduce the project items' font-size from 18px to 1rem
- Add a text-overflow to the items, with a tooltip to display the cropped title, instead of the horizontal scroll

Current version:
image

Adjusted version:
image

fsmaia:
This fixes #581.

It reduces project items' font size to 16px, reduces their left padding to 16px, and removes `white-space: no-wrap` from the sidebar, so large project names are displayed properly, both in the project title and in the sidebar item.

Previously, I thought of ellipsis with some kind of tooltip, but I see no problem with this implementation too. Anyway, I'm open to suggestions.

Additionally, there is no horizontal scroll on the sidebar anymore.

To test this, just add some very large project names, as in the following screenshots:

image

image

### DIFF
--- a/packages/dashboard/src/components/layout/sidebar.tsx
+++ b/packages/dashboard/src/components/layout/sidebar.tsx
@@ -126,7 +126,6 @@ const Drawer = styled(MuiDrawer, {
   return {
     width: DRAWER_WIDTH,
     flexShrink: 0,
-    whiteSpace: 'nowrap',
     boxSizing: 'border-box',
     ...(open && {
       ...openedMixin(theme),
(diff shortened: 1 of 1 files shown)


fsmaia:
In my humble opinion, I would also reduce the drawer header to match the toolbar's height (currently the drawer header is locked to 100px), I thought it be a little fancier:
image

PR #610, opened 2022-07-25
Fix word break overflow

fsmaia:
Fixes an edge case of #608.

It forces `word-wrap: break-all` on sidebar items and dashboard project list item card, to avoid horizontal scroll on very large words. To test it, just add a project with a very long word, or even a long word right before the first row ends.

Note: if you wish I can add a polyfill, but IMHO is not necessary:
https://caniuse.com/?search=word-break

Before:

image

image

After:

image

image

### DIFF
--- a/packages/dashboard/src/components/layout/sidebar.tsx
+++ b/packages/dashboard/src/components/layout/sidebar.tsx
@@ -174,6 +174,7 @@ export const ProjectListMenu: ProjectListMenuType = ({
                   fontSize: 16,
                   color: 'text.primary',
                   sx: { opacity: 0.6 },
+                  style: { wordBreak: 'break-all' },
                 }}
               />
             </ListItemButton>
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/layout/sidebar.tsx
```diff
@@ -174,6 +174,7 @@ export const ProjectListMenu: ProjectListMenuType = ({
                   fontSize: 16,
                   color: 'text.primary',
                   sx: { opacity: 0.6 },
+                  style: { wordBreak: 'break-all' },
```
fsmaia, line 177:
```suggestion
                  style: { wordBreak: 'break-word' },
```

packages/dashboard/src/project/projectListItem.tsx
```diff
@@ -52,6 +52,7 @@ export function ProjectListItem({ project }: ProjectListItemProps) {
                 primary={decodeURIComponent(project.projectId)}
                 primaryTypographyProps={{
                   variant: 'body1',
+                  style: { wordBreak: 'break-all' },
```
fsmaia, line 55:
```suggestion
                  style: { wordBreak: 'break-word' },
```

Issue #613, opened 2022-08-04
Off-by-one error in sorry-cypress causes browser to be closed prematurely in Cypress >=10


alyssa-glean:
Please note - issues that are not following the template will be closed automatically:

- [x] I have verified that all sorry-cypress services are running and accessible
- [x] I have read documentation and found no answer to my problem
- [x] I have searched issues / discussions and found to answer to my problem
- [ ] I have read [Minio Configutation guide](https://docs.sorry-cypress.dev/configuration/director-configuration/minio-configuration) - for issues related to Screenshots / Videos uploads for Minio - N/A
- [ ] I have activated the debug mode `DEBUG=cypress:server:api cy2 ....` and have the network error - N/A

## Summary

When using the `sorry-cypress` director, there is an off-by-one error in the test index that's reported in the logging. The first spec gets labelled "2 of N", and this continues right up until the final spec which gets reported as "N+1 of N". Until Cypress 10, this was just an inconsequential quirk, but is now causing actual problems for us having bumped to Cypress 10.3.0.

Previously, Cypress used to launch a new browser per test spec. As of Cypress 10, though, it launches one browser at the start of the run and opens a new tab for each new spec (see [this issue comment](https://github.com/cypress-io/cypress/issues/2951#issuecomment-1188491798)). I believe that it then closes the browser when the "last" spec has been run.

What this means in practice is that if a runner picks up the two final specs of a run, you end up with an error. After it runs the first of these the browser is closed (because sorry-cypress says that it's e.g. "18 of 18"). The final spec then tries to run, and errors out because the browser is no longer available.

N.B. I found this issue which is the same thing: https://github.com/sorry-cypress/sorry-cypress/issues/543. However, the title is misleading as to what's actually happening, and it was closed for inactivity. So I thought it best to raise a new issue.

## How to reproduce

I can replicate this 100% of the time by using `sorry-cypress` with a single runner and Cypress 10.3.0. With multiple runners it only happens sometimes (because runs where specs N-1 and N are picked up by different runners are unaffected). Taking `sorry-cypress` out of the equation removes the problem entirely.

## Relevant logs / screenshots

See the following logs from the end of a run using a single runner and sorry-cypress, showing the final two specs and what happens:

```
[2022-08-03T16:02:26.477Z] ────────────────────────────────────────────────────────────────────────────────────────────────────
[2022-08-03T16:02:26.477Z]                                                                                                     
[...]

PR #615, opened 2022-08-04
Fix off-by-one error in the in-memory director

alyssa-glean:
Fixes https://github.com/sorry-cypress/sorry-cypress/issues/613

I have also: 

 - Reverted the change made to the `tests.yml` in https://github.com/sorry-cypress/sorry-cypress/pull/427. This change unintentionally led to the actions *never running* - the last time this action ran in this repo was August last year... :sweat_smile: 
 - Fixed the `hooksReporter.test.ts` suite in CI by building the logger package and updating the fixtures (I guess this broke with no one noticing because CI hasn't been running)

You can see a successful CI run in my forked repo [here](https://github.com/alyssa-glean/sorry-cypress/actions/runs/2797866730)

If any of my changes to the test fixtures look wrong/unintended, then I guess that's another bug that needs fixing. I've avoided scrutinising them too much as I had already gone quite off-piste from where I began with the PR!

Finally, I noticed that there is similar logic in the mongo implementation - see this code in `run.controller.ts`:

```
try {
    await setSpecClaimed(runId, groupId, spec.instanceId, machineId);
    await createInstance({
      runId,
      projectId: run.meta.projectId,
      instanceId: spec.instanceId,
      groupId,
      spec: spec.spec,
      cypressVersion,
    });
    return {
      instance: spec,
      claimedInstances: getClaimedSpecs(run, groupId).length + 1,
      totalInstances: getSpecsForGroup(run, groupId).length,
      projectId: run.meta.projectId,
    };
```

It *looks* like the +1 might be unnecessary here, too, but it depends on whether the call to `setSpecClaimed` updates the in-memory `run` object or not (at a glance, it might just update mongodb). We don't use the mongo implementation at my company so I don't know if it's actually broken or not, and have therefore decided not to change it. There are also no existing tests to crib from and I suspected that writing one would be more complicated and involve mocking out mongo etc.


agoldis:
@all-contributors please add @alyssa-glean for coding

agoldis:
https://github.com/all-contributors please add @alyssa-glean for coding

agoldis:
@all-contributors please add @alyssa-glean for coding

agoldis:
@alyssa-glean Thanks for the very detailed description and the extra work on the GH actions!

PR #650, opened 2022-09-20
fix: director returning 500 using gitlab job retries

bjartur20:
Bug reported here on the original [PR](https://github.com/sorry-cypress/sorry-cypress/pull/641#issuecomment-1251988954)

The cause of the bug was as @vinhoe suggested, the getFailedSpecs function was throwing an error when specs had already in reset.

[Bug ticket](https://github.com/sorry-cypress/sorry-cypress/issues/652#issue-1380911426)

### DIFF
--- a/packages/director/src/execution/utils.ts
+++ b/packages/director/src/execution/utils.ts
@@ -10,7 +10,7 @@ export const getFirstUnclaimedSpec = (run: Run, groupId: string) =>
   getSpecsForGroup(run, groupId).find((s) => !s.claimedAt);
 export const getFailedSpecs = (run: Run, groupId: string) =>
   getSpecsForGroup(run, groupId).filter((s) =>
-    s.results ? s.results?.stats.failures + s.results.stats.skipped > 0 : false
+    s.results ? s.results.stats.failures + s.results.stats.skipped > 0 : false
   );
 
 interface GetNewSpecsForGroupParams {
--- a/packages/director/src/execution/utils.ts
+++ b/packages/director/src/execution/utils.ts
@@ -9,8 +9,8 @@ export const getClaimedSpecs = (run: Run, groupId: string) =>
 export const getFirstUnclaimedSpec = (run: Run, groupId: string) =>
   getSpecsForGroup(run, groupId).find((s) => !s.claimedAt);
 export const getFailedSpecs = (run: Run, groupId: string) =>
-  getSpecsForGroup(run, groupId).filter(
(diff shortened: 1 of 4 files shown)


packages/director/src/execution/utils.ts
```diff
@@ -9,7 +9,9 @@ export const getClaimedSpecs = (run: Run, groupId: string) =>
 export const getFirstUnclaimedSpec = (run: Run, groupId: string) =>
   getSpecsForGroup(run, groupId).find((s) => !s.claimedAt);
 export const getFailedSpecs = (run: Run, groupId: string) =>
-  getSpecsForGroup(run, groupId).filter((s) => s.results!.stats.failures > 0);
```
mathpaquette, line 12: @bjartur20 did you try simply replacing ! by ?
```
getSpecsForGroup(run, groupId).filter((s) => s.results?.stats.failures > 0);
```

mathpaquette: @bjartur20 hello?

bjartur20: Yeah sorry for the late reply.
I tried that but that still resolves in a type error since s.results is possibly undefined.
I thank that my fix should cover this, since now it'll only check for the results if they are defined.

mathpaquette: @bjartur20 could you please provide steps to reproduce ?

bjartur20: I created a bug issue which includes steps to reproduce: https://github.com/sorry-cypress/sorry-cypress/issues/652#issue-1380911426

packages/director/src/execution/utils.ts
```diff
@@ -9,7 +9,9 @@ export const getClaimedSpecs = (run: Run, groupId: string) =>
...
+  getSpecsForGroup(run, groupId).filter(
```
agoldis, line 12: Technically `skipped` are also failed, cypress just isn't running those tests and marks them as skipped.
```ts
getSpecsForGroup(run, groupId).filter((s) =>
    s.results ? s.results.stats.failures + s.results.stats.skipped > 0 : false
  );
```

agoldis (changes requested):
- [`run.model.ts`](https://github.com/sorry-cypress/sorry-cypress/blob/master/packages/director/src/execution/mongo/runs/run.model.ts#L286) also needs a better type guards to avoid using `!`

PR #654, opened 2022-09-22
feat: make USER an ARG so containers based on this one can switch between root and $USER

rrauenza:
For bug fixes:

- [x ] Add a reference to the original issue - #645
- [x ] Add a test and / or some kind of instructions to verify the fix  - none needed


agoldis:
@allcontributors please add @rrauenza for code

PR #671, opened 2022-10-19
changed gitlab_job_retries to general retries

raftx24:
I needed retries in bitbucket so I tried to change `GITLAB_JOB_RETRIES` to a general option so we can use it everywhere. I renamed the flag and I changed two conditions, now it works well in bitbucket too.


agoldis:
@raftx24 Seems like this one is a bit more complex than it seems:

- does it support grouping?
- what about other CI providers?

PR #688, opened 2022-11-18
Add new app authentication mechanism for github hooks

Spea:

### Use case

Currently it is only possible to use a token from a specific user for the github hook. This can be solved in two ways:

* Use token from my own account -> not so good, as I could leave the company at any time
* Create a new machine user -> Increases the amount of users you have to pay for

To mitigate these issues, I added the app authentication mechanism to the github hook. 
The provided solution is also backward compatible.

### Octokit dependency

I added the offical github [octokit.js](https://github.com/octokit/octokit.js) dependency, as this library makes it quite easy to support different authentication mechanism for github. Let me know if that is ok for you. I could also implement the different API calls with axio, though it will be more work.

#### Token authentication (default)

image

#### App authentication mechanism

image

image


agoldis:
@all-contributors please add @Spea for code

agoldis:
Thanks a lot @Spea, nicely done!

PR #695, opened 2022-11-22
Fixing  TS conflicts

agoldis:
@Spea not sure why #688 didn't invoke GH actions type checks, but the PR caused a bunch of type and build errors. I have resolved most of them, except [this](https://github.com/sorry-cypress/sorry-cypress/actions/runs/3520971966/jobs/5902372604).

Can you please help?

PR #696, opened 2022-11-22
Only install necessary octokit packages

Spea:
Requiring all of `octokit` is a bit overkill and installs a lot of packages that are actually not used. Only installing `@octokit/rest` and `@octokit/auth-app` is enough for our use case.

Fixes the linting error mentioned in #695

@agoldis Again sorry for the inconvenience, should have been more thorough

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,7 @@ name: Lint and test
 
 on:
   push:
+  pull_request:
 
 jobs:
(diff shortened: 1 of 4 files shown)


Spea:
@agoldis Seems like the linting and tests were not triggered, because of the missing `pull_request` event in the workflow file. I added it via 4e152688486fafb22ba1bb777e06bd9255627b0f

PR #699, opened 2022-11-24
Add possibility to chooses to use https or http for links

EHoglid:
For new features:

In Sorry Cypress you always assume that the connection to GitLab, GitHub or other web based git platforms are https:// but that is not always the case. If you for example have a self maintained environment then it is possible that this is not configured and you only use http://.
I have configured a variable USE_SSL_FOR_LINKS that can be set the director part when creating the director in docker.
'true' = use https:// 
'fasle' = use http://

image

Also I discovered that in the dashboard it do the same thing as in the director, converting git@ to https:// so I removed that function handleSshUrl.
The getRemoteOrigin() function now takes a argument if it should return http or https.

By setting the default value to 'true' this should be backwards-compatible

Created a own docker and pulled the image and tested it on my computer. (ehoglid/sorry-cypress-....)
All test passed and extended the test for getRemoteOrigin() to test the extra argument and check if the return is OK.

USE_SSL_FOR_LINKS: 'false'
image

USE_SSL_FOR_LINKS: 'true'
image

No argument in docker, default should be https (backwards-compatible)
image


packages/director/src/execution/__tests__/utils.test.ts
```diff
@@ -6,26 +6,42 @@ describe('Driver utility functions', () => {
...
+    {
+      description: 'ssh protocol',
+      remoteOrigin: 'ssh@gitlab.com:group/project.git',
+      expected: 'http://gitlab.com/group/project.git',
+      useSSH: 'false',
+    },
+  ].forEach(({ description, remoteOrigin, expected, useSSH }) => {
     it(`parses remoteOrigin for ${description} correctly`, () => {
-      const href = getRemoteOrigin(remoteOrigin);
+      const href = getRemoteOrigin(remoteOrigin, useSSH);
```
agoldis, line 44: useSSL?

packages/dashboard/src/lib/github.ts
```diff
@@ -1,8 +1,5 @@
 export const getGithubCommitURL = (repo: string, sha: string) =>
-  handleSshURL(repo.replace(/.git$/, '') + `/commit/${sha}`);
+  repo.replace(/.git$/, '') + `/commit/${sha}`;
 
 export const getGithubBranchURL = (repo: string, branch: string) =>
-  handleSshURL(repo.replace(/.git$/, '') + `/tree/${branch}`);
-
-export const handleSshURL = (sshUrl: string) =>
```
agoldis, line 7: hm.. why was it removed?

agoldis (commented):
- [ ] Let's rename the variable to "USE_HTTPS_FOR_GIT_ORIGIN"?
- [ ] How about moving the use of `USE_SSL_FOR_LINKS` to `packages/director/src/execution/utils.ts`
- [ ] Please update the documentation at https://github.com/sorry-cypress/gitbook

packages/dashboard/src/lib/github.ts
EHoglid: What I understand and what I can see in the code this is now handled only in the director converting git URL to a HTTP/HTTPS URL. The dashboard will only get the the Branch URL as either https:// or http://

Example:
director service will convert this data 
from:
git@ts-gitlab.sorrycypress.lan:web-dev/testing_something.git 

to:
 https://ts-gitlab.sorrycypress.lan/web-dev/testing_something.git
 
 So why do this again in the dashboard?

mathpaquette:
@EHoglid I dont understand we why need an option for it ? Technically it should be automatically handled no ?

EHoglid:
@mathpaquette 
It depends, normally you can have a settings that redirect http to https if you support https. But in my case our GitLab isn't setup with https, then you can't handle https requests at all.
Second, I have never heard that you redirect a https request to http, that would be a security issue.
Therefor it would be good to choose if it should be http or https. 
Of course https is the preferred solution because it is a encrypted connection but in our case it is a local server in a closed environment, therefore it has never been setup.

But I do not see any issue of supporting both? https is the default value anyway, so it won't affect any users if this is implemented.

CloneOfAlex (approved):

Issue #519, opened 2022-01-08
Screenshots uploaded to AWS S3 have wrong Content-Type , hence Chrome downloads image instead of displaying inline


mmtmohit:
## Summary

The content type metadata of uploaded screenshot images is 'binary/octet-stream' (as seen in Management Console) while It should be 'image/png' so that when I click on image displayed inline to the failed test on the dashboard , browser can open/display inline in same/new tab instead. of force download.

I checked the logs of director service and can see the upload url has query param Content-Type being sent as 'image/png' , yet when I checked metadata for the S3 object in AWS management console, it shows 'binary/octet-stream' strange enough.
Question is why the s3 sdk is not putting the right content-type of the image.

## How to reproduce

1. Deployed latest docker versions of director, dashboard and api service to herokuapp
2. Added new S3 bucket manually in AWS, new IAM user with below. policy :
    {
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "VisualEditor0",
            "Effect": "Allow",
            "Action": [
                "s3:Put*"
            ],
            "Resource": [
                "arn:aws:s3:::xxxxxxbucket-east/*"
            ]
        }
    ]
}
3. Added Cloudfront distribution in front of s3 origin and added OAI policy to S3 bucket
4. Ran tests pointing to director service

## Relevant logs / screenshots
AWS s3 console - object metadata :
<img width="1560" alt="Screen Shot 2022-01-08 at 11 54 09 AM" src="https://user-images.githubusercontent.com/11607818/148634228-04d55c65-be9d-42b2-bdc4-ace1b225b31d.png">

dashboard failed test screenshot :
<img width="1541" alt="Screen Shot 2022-01-08 at 11 59 07 AM" src="https://user-images.githubusercontent.com/11607818/148634401-86a56cd8-cad5-4a78-b0f6-4b4ed8547bd8.png">
Exact Issue - When the above screenshot image is clicked to open in browser, it gets force downloaded in place of. opening inline

- relevant logs from director service:

[...]

PR #700, opened 2022-11-25
feat: Add screenshot zoom in/out on preview functionality

matrunchyk:
New feature:

- [x] Discussed here: https://github.com/sorry-cypress/sorry-cypress/issues/519
- [x] Use case: when I click on a screenshot, I should be able to preview & zoom instead of downloading it.
- [x] Backward compatibility is not relevant to this fix (the UI will work with older versions of other packages)
- [x] See [video recording](https://github.com/sorry-cypress/sorry-cypress/issues/519#issuecomment-1327914582) for demo

### DIFF
--- a/packages/dashboard/package.json
+++ b/packages/dashboard/package.json
@@ -50,6 +50,7 @@
     "react-color": "^2.19.3",
     "react-dom": "^17.0.2",
     "react-hook-form": "^6.15.8",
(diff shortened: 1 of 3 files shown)


agoldis (approved):

maximilianschmid:
Thank you, a long awaited improvement!

PR #741, opened 2023-02-11
Turborepo

nmengual:

## References

## Use case

Added turborepo to be able to launch tasks in parallel, cache steps and get a faster ci/cd and development environment.

Scripts are also configured to take into account dependencies between projects and between scripts (for example the test script does a build if necessary before running).


agoldis:
@all-contributors please add @nmengual for code

PR #759, opened 2023-03-21
Start building Arm64 Docker image

### LINKED ISSUE #734: Multi Architecture Docker Images
- [x] I have searched issues for similar feature requests

## What is this feature about (1 sentence)

I want to be able to run sorry-cypress on ARM instances and currently only AMD is supported.

## Why is it needed? What is the value? For whom do we build it?

ARM is more performant when running Sorry Cypress on AWS and is cheaper.

## High-level feature overview

Build Docker Images with `docker buildx` with both AMD/ARM types.

pavlospt:
## References

- [ ] I have updated the [documentation](https://github.com/sorry-cypress/gitbook).
- [ ] I have added included automated tests
- [x] Original issue / feature request / discussion: `https://github.com/sorry-cypress/sorry-cypress/issues/734`

## Use case

Addition of Arm64 Docker images, in order to be able to deploy it in Graviton based nodes.

## Example

-

Closes #734

### DIFF
--- a/scripts/release-dockerhub.sh
+++ b/scripts/release-dockerhub.sh
@@ -43,26 +43,22 @@ function getTagsArg() {
   done
 }
 
-function dockerBuild() {
-  echo 🔨 Building ${2} from ${1}: docker build --file ${1}/Dockerfile $(getTagsArg ${2})
+function dockerBuildxSetup() {
+  echo Setting Up Docker Buildx
   echo ========================
-  docker build --file ${1}/Dockerfile $(getTagsArg ${2}) .
+  docker buildx use default
   echo ========================
   echo ✅ Build completed ${2} from ${1} 
 }
 
-function dockerPush() {
-  for TAG in ${TAGS}
-  do
-    echo 💾 Pushing to remote: docker push ${1}:${TAG}
-    echo ========================
-    docker push "${1}:${TAG}"
-    echo ========================
-    echo ✅ Pushed "${1}:${TAG}"
-  done
+function dockerBuildAndPush() {
+  echo 🔨 Building ${2} from ${1}: docker buildx --file ${1}/Dockerfile --platform=linux/arm64,linux/amd64 $(getTagsArg ${2}) --provenance=false --push
+  echo ========================
+  docker buildx --file ${1}/Dockerfile --platform=linux/arm64,linux/amd64 $(getTagsArg ${2}) --provenance=false --push .
+  echo ========================
+  echo ✅ Build \& push completed ${2} from ${1} 
 }
 
-# ./scripts/release-dockerhub.sh -t cypress-v5
 while getopts t:s: flag
 do
     case "${flag}" in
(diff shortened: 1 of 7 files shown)


scripts/release-dockerhub.sh
```diff
@@ -43,10 +43,18 @@ function getTagsArg() {
...
+  docker buildx create --platform=linux/arm64,linux/amd64 --use
+  echo ========================
+  echo ✅ Build completed ${2} from ${1} 
+}
+
 function dockerBuild() {
   echo 🔨 Building ${2} from ${1}: docker build --file ${1}/Dockerfile $(getTagsArg ${2})
   echo ========================
-  docker build --file ${1}/Dockerfile $(getTagsArg ${2}) .
+  docker buildx --file ${1}/Dockerfile --platform=linux/arm64,linux/amd64 $(getTagsArg ${2}) .
```
dlavrenuek, line 57: Using `docker buildx build ...` and afterwards `docker push ...` will not work. The images + manifest for multiarch docker images are not added to your local docker registry but are only "temporary" in the buildx environment. Instead you can use [`--push`](https://docs.docker.com/engine/reference/commandline/buildx_build/#push) to directly build+push and remove the `dockerPush` step.

Also note that due to the new feature `SLSA Provenance attestation`, which is enabled by default since January's 0.10.x release of `docker buildx` (see [release notes](https://github.com/docker/buildx/releases/tag/v0.10.0)), the produced image manifest will be of media type `application/vnd.docker.distribution.manifest.list.v1+json` instead of `application/vnd.docker.distribution.manifest.list.v2+json`. This causes problems on some CI platforms, such as Google Cloud or Jfrog artifactory. I would recommend using `--provenance=false` for compatibility reasons.

dlavrenuek (changes requested):
Thank you for the changes @pavlospt 🎉 . The build + push process is not easy to test since you don't have access to push these images to docker hub, so I have added a comment regarding the challenges I have encountered with the same concept.

scripts/release-dockerhub.sh
pavlospt: You are right, indeed. The way you have described it, is how we do it for our private repos, but was not sure of this project's structure so I did not make any more changes to the push of the Docker image. I will update the PR with your comments 🙏🏻

pavlospt: @dlavrenuek PR updated!

adferrand:
I should have looked at the PR list before starting by duplicate of this work here :) (#777 for reference)

Still, I would recommend that you apply on this PR my modifications about the Dockerfiles. Indeed, since the runtime is Node, you do not really need to build twice the project as part of the `build` stage, and you certainly want to avoid to do this using the QEMU emulation that is 10x slower than running on the native architecture. My changes are pinning the `build` phase to `amd64` architecture which is used by GitHub Action workers.

Cheers !

der-eismann:
Is there any update here? We could really need an ARM-based image here :slightly_smiling_face:

agoldis:
@allcontributors please add @pavlospt for code

PR #775, opened 2023-04-20
chore: bitbucket cloud url processing for commit and PR links

FilipSwiatczak:

## References

- [ X] I have updated the [documentation](https://github.com/sorry-cypress/gitbook). PR link `<here>`
- [ X] I have added included automated tests
- [ X] I acknowledge that I have tested that the change is backwards-compatible
- [ X] Original issue / feature request / discussion: `<here>`

## Use case

Dashboard would use a separate function to treat Bitbucket Cloud links to Commits and PRs. 
`https://bitbucket.org/<org>/<repo>/commits/<hash>` <-- commit**S**
`https://bitbucket.org/<org>/<repo>/pull-requests/76` <-- pull-requests instead of tree, and PR- removed

I'm including the Bitbucket update in this PR. There is also new lib/repository.ts that detects if github/bitbucket is being used and returns correct implementation to the commit.tsx
This is basic but easy to read and understand and hopefully easy to extend for others like GitLabs or Bitbucket server. 

Documentation: this is an invisible improvement and I didn't see any section on gitbook that would warrant an addition, please correct if I'm wrong.
Tests: I would have written unit tests but the Dashboard does not look to have any test framework added (like jest) and preferred not to introduce too much change.

Bitbucket url to ssh regex should be working:
image

We are looking forward to using Sorry Cypress with Bitbucket cloud. Thanks!

### DIFF
--- a/packages/dashboard/src/lib/repository.ts
+++ b/packages/dashboard/src/lib/repository.ts
@@ -1,0 +1,25 @@
+import {
+  getBitbucketCloudBranchURL,
+  getBitbucketCloudCommitURL,
+  handleBitbucketCloudSshURL,
+} from '@sorry-cypress/dashboard/lib/bitbucketCloud';
+import {
+  getGithubBranchURL,
+  getGithubCommitURL,
+  handleGithubSshURL,
+} from '@sorry-cypress/dashboard/lib/github';
+
+export const getCommitURL = (repo: string, sha: string) =>
+  repo.includes('bitbucket.org')
+    ? getBitbucketCloudCommitURL(repo, sha)
+    : getGithubCommitURL(repo, sha); // github is default
+
+export const getBranchURL = (repo: string, branch: string) =>
+  repo.includes('bitbucket.org')
+    ? getBitbucketCloudBranchURL(repo, branch)
+    : getGithubBranchURL(repo, branch); // github is default
+
+export const handleSshURL = (sshUrl: string) =>
+  sshUrl.includes('bitbucket.org')
+    ? handleBitbucketCloudSshURL(sshUrl)
+    : handleGithubSshURL(sshUrl); // github is default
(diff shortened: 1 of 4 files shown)


mathpaquette:
@FilipSwiatczak what about https:// urls?

FilipSwiatczak:
> @FilipSwiatczak what about https:// urls?

hi @mathpaquette , the bitbucket ssh url is almost identical to the github one with the distinction of having **.git** at the end, therefore the:
```
export const handleBitbucketCloudSshURL = (sshUrl: string) =>
  sshUrl.replace(/git@([\w.]+):(.+).git/, 'https://$1/$2');
```

Is a word for word copy of the previous handleSshUrl with the exception of regex including above **.git**. 
Those are really cosmetic changes. Does that make sense? Cheers

packages/dashboard/src/lib/repository.ts
```diff
@@ -0,0 +1,25 @@
...
+  handleBitbucketCloudSshURL,
+} from '@sorry-cypress/dashboard/lib/bitbucketCloud';
+import {
+  getGithubBranchURL,
+  getGithubCommitURL,
+  handleGithubSshURL,
+} from '@sorry-cypress/dashboard/lib/github';
+
+export const getCommitURL = (repo: string, sha: string) =>
+  repo.includes('bitbucket.org')
```
mathpaquette, line 13: I was expecting this code not to be specific to bitbucket since it's in repository.ts file

mathpaquette: also, just a bit sad that it will just work with bitbucket cloud. pretty sure its the same logic needed for on prem version

mathpaquette: @FilipSwiatczak yeah but a cleaner way of doing it would be, each implementation should provide one canHandle method and we create an array with all supported implementation (in repository.ts) and we iterate through them til we find supported one.

@agoldis are you okay with those changes ?

agoldis:
@mathpaquette @FilipSwiatczak thanks for reviewing it, if you guys think it's a good PR and no other comments please feel free to proceed

PR #796, opened 2023-06-16
Support signed S3 download URLs

crux-capacitor:

## References

- [ ] I have added inclued automated tests
- [ ] I acknowledge that I have tested that the change is backwards-compatible
- [ ] Original issue / feature request / discussion: `<here>`

## Use case

> For security reasons, the S3 screenshots bucket in my deployment was required to be encrypted with a CMK. This started leading to 400 Bad Request responses from S3 when the browser would try to download the video file for a test.

To get around this, SC should return signed download URLs.

## Example

> Submit screenshots / outputs / tests together with the PR.

### DIFF
--- a/cloudformation/sorry-cypress.yml
+++ b/cloudformation/sorry-cypress.yml
@@ -313,6 +313,7 @@ Resources:
     Type: AWS::S3::Bucket
     Properties:
       BucketName: !Join ['-', [!Ref 'AWS::StackName', 'test-results']]
+      AccessControl: PublicRead
       CorsConfiguration:
         CorsRules:
           - AllowedHeaders:
--- a/cloudformation/sorry-cypress.yml
+++ b/cloudformation/sorry-cypress.yml
@@ -313,7 +313,6 @@ Resources:
     Type: AWS::S3::Bucket
     Properties:
       BucketName: !Join ['-', [!Ref 'AWS::StackName', 'test-results']]
-      AccessControl: PublicRead
       CorsConfiguration:
         CorsRules:
           - AllowedHeaders:
--- a/packages/dashboard/src/instance/instanceDetails.tsx
+++ b/packages/dashboard/src/instance/instanceDetails.tsx
@@ -3,6 +3,21 @@ import ChevronRightIcon from '@mui/icons-material/ChevronRight';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import TreeItem from '@mui/lab/TreeItem';
 import TreeView from '@mui/lab/TreeView';
+import aws from 'aws-sdk';
+import { AssetDownloadInstruction } from '@sorry-cypress/common';
+import {
+  S3_ACL,
+  S3_BUCKET,
+  S3_REGION,
+  S3_PRESIGN_DOWNLOAD_URLS,
+  S3_FORCE_PATH_STYLE,
+  S3_READ_URL_PREFIX,
+  S3_IMAGE_KEY_PREFIX,
+  S3_VIDEO_KEY_PREFIX,
+  S3_DOWNLOAD_EXPIRY_SECONDS,
+} from '../../../director/src/screenshots/s3/config';
+import * as s3 from '../../../director/src/screenshots/s3/s3';
+import { S3SignedDownloadResult } from '../../../director/src/screenshots/s3/types';
 import {
   Alert,
   AlertTitle,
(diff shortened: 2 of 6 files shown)


cloudformation/sorry-cypress.yml
```diff
@@ -313,7 +313,6 @@ Resources:
     Type: AWS::S3::Bucket
     Properties:
       BucketName: !Join ['-', [!Ref 'AWS::StackName', 'test-results']]
-      AccessControl: PublicRead
```
crux-capacitor, line 316: This resolves a breaking change due to AWS updating allowed default settings on new buckets.

packages/dashboard/src/instance/instanceDetails.tsx
```diff
@@ -359,6 +374,13 @@ function getTreeOfNavigationItems(instance: GetInstanceQuery['instance']) {
 
   // if instance has video add it as the first item of the navigation tree
   if (instance?.results?.videoUrl) {
+
+    if (S3_PRESIGN_DOWNLOAD_URLS === true) {
+      const key = instance.results.videoUrl.split('/').pop()  // split url and get last item to get object key
+      const signedDownloadUrl = s3.getDownloadUrl(key)
+      instance.results.videoUrl = signedDownloadUrl
```
crux-capacitor, line 381: This `split` doesn't work - it says `Property 'split' does not exist on type 'string'.`. How can I make this work?

packages/director/src/screenshots/s3/config.ts
```diff
@@ -5,7 +5,12 @@ export const S3_FORCE_PATH_STYLE = process.env.S3_FORCE_PATH_STYLE || false;
 export const S3_READ_URL_PREFIX = process.env.S3_READ_URL_PREFIX || undefined;
 export const S3_IMAGE_KEY_PREFIX = process.env.S3_IMAGE_KEY_PREFIX || undefined;
 export const S3_VIDEO_KEY_PREFIX = process.env.S3_VIDEO_KEY_PREFIX || undefined;
-export const UPLOAD_EXPIRY_SECONDS = parseInt(
-  process.env.UPLOAD_EXPIRY_SECONDS || '90',
+export const S3_PRESIGN_DOWNLOAD_URLS = process.env.S3_PRESIGN_DOWNLOAD_URLS || false;
```
crux-capacitor, line 8: This is the new bool variable that triggers generating signed download URLs.

cloudformation/sorry-cypress.yml
agoldis: Sorry for my ignorance, what do you mean here?

packages/dashboard/src/instance/instanceDetails.tsx
```diff
@@ -3,6 +3,21 @@ import ChevronRightIcon from '@mui/icons-material/ChevronRight';
...
+  S3_ACL,
+  S3_BUCKET,
+  S3_REGION,
+  S3_PRESIGN_DOWNLOAD_URLS,
+  S3_FORCE_PATH_STYLE,
+  S3_READ_URL_PREFIX,
+  S3_IMAGE_KEY_PREFIX,
+  S3_VIDEO_KEY_PREFIX,
+  S3_DOWNLOAD_EXPIRY_SECONDS,
+} from '../../../director/src/screenshots/s3/config';
```
agoldis, line 18: I am not sure you can use the config values (in runtime) from other package. One approach would be to generate signed URLs on the API side

packages/director/src/screenshots/s3/config.ts
agoldis: The previous variable `UPLOAD_EXPIRY_SECONDS` was being used in minio driver as well, we need to preserve it

packages/director/src/screenshots/s3/s3.ts
```diff
@@ -74,5 +82,47 @@ export const getVideoUploadUrl = async (
...
+  key,
+  Expires = S3_DOWNLOAD_EXPIRY_SECONDS,
+}: GetDownloadURLParams): Promise<S3SignedDownloadResult> => {
+  const s3Params = {
+    Bucket: S3_BUCKET,
+    Key: key,
+    Expires
+  };
+
+  return new Promise((resolve, reject) => {
```
agoldis, line 99: The signed download URL will be persisted to the DB with the expired signature. I think the signing should be done on the API each time we retrieve the url

agoldis (changes requested):
Seems like it needs more work - please see the inline comments

agoldis (changes requested):
Seems like it needs more work - please see the inline comments

agoldis:
@crux-capacitor

Issue #805, opened 2023-06-30
Make director/screenshots/s3 package compatible with other providers


p-8-z:
- [x] I have searched issues for similar feature requests

## Why is it needed? What is the value? For whom do we build it?

Make director/screenshots/s3 package compatible with other providers.
Make non-public Buckets accessible even for downloading (pre-signed download urls). 

## High-level feature overview

- updated AWS library (fix AWS warnings)
- url signing for upload and download
- make director/screenshots/s3 package compatible with other providers (tested with Oracle OCI S3)

p-8-z:
Implemented:
#806

actuallymentor:
Is this a good place to chime in and ask for file storage on the director's server? Or is that beyond the scope of this issue?

agoldis:
That's beyond I am afraid

Rohmilchkaese:
Sorry, but that doesn't seem true according to: [sorry-cypress/packages/director/src/screenshots/s3/s3.ts](https://github.com/sorry-cypress/sorry-cypress/blob/87e8f814e1c48132ad8620f3338007d8fea1ffe0/packages/director/src/screenshots/s3/s3.ts#L17)

```
import { AssetUploadInstruction } from '@sorry-cypress/common';
import aws from 'aws-sdk';
import { sanitizeS3KeyPrefix } from '../utils/';
import {
  S3_ACL,
  S3_BUCKET,
  S3_FORCE_PATH_STYLE,
  S3_IMAGE_KEY_PREFIX,
  S3_READ_URL_PREFIX,
  S3_REGION,
  S3_VIDEO_KEY_PREFIX,
  UPLOAD_EXPIRY_SECONDS,
} from './config';
import { S3SignedUploadResult } from './types';

const BUCKET_URL = S3_FORCE_PATH_STYLE
  ? `https://s3.amazonaws.com/${S3_BUCKET}`
  : `https://${S3_BUCKET}.s3.amazonaws.com`;

```

I'm not a coder, but I dont see a possibility to set different URL from AWS here, am I right?

Rohmilchkaese:
@agoldis Any Updates to https://github.com/sorry-cypress/sorry-cypress/pull/806 ?

Rohmilchkaese:
Is there something we can do as company to push this? We love sorry cypress and would love to have the capability to push to an self hosted s3 for eg.

agoldis:
@Rohmilchkaese you can use self-hosted s3 via minio driver:
[...]

PR #806, opened 2023-06-30
feature(director): s3 update, make it compatible with other providers

p-8-z:
I've update aws sdk to latest version and made director/screenshots/s3 package compatible with other providers.

## References

- [x] I have updated the [documentation](https://github.com/sorry-cypress/gitbook). PR link [here](https://github.com/sorry-cypress/gitbook/pull/26)
- [ ] I have added included automated tests
- [ ] I acknowledge that I have tested that the change is backwards-compatible
- [x] Original issue / feature request / discussion: [here](https://github.com/sorry-cypress/sorry-cypress/issues/805)

## Use case

-  director/screenshots/s3 package compatible with other providers.

## Example

Tested against Oracle S3 compatibility API

```
  director:
    build:
      dockerfile: packages/director/Dockerfile
      context: .

    image: agoldis/sorry-cypress-director:${VERSION:-latest}
    environment:
      DASHBOARD_URL: http://localhost:8080
      MONGODB_URI: 'mongodb://mongo:27017'
      MONGODB_DATABASE: 'sorry-cypress'
      EXECUTION_DRIVER: '../execution/mongo/driver'
      SCREENSHOTS_DRIVER: '../screenshots/s3.driver'
      GITLAB_JOB_RETRIES: 'false'
      PROBE_LOGGER: "false"
      #Bucket
      S3_URL: https://namespace.compat.objectstorage.region.oraclecloud.com
      S3_REGION: eu-frankfurt-1
      S3_BUCKET: sorry-cypress
      AWS_ACCESS_KEY_ID: 'key'
      AWS_SECRET_ACCESS_KEY: 'secret'
    ports:
      - 1234:1234
    depends_on:
      - mongo
```


packages/director/src/screenshots/s3/s3.ts
```diff
@@ -1,78 +1,64 @@
...
-  const s3Params = {
-    Bucket: S3_BUCKET,
-    Key: key,
-    Expires,
-    ContentType,
-    ACL: S3_ACL,
-  };
+    const url = parseUrl(`${S3_URL}/${S3_BUCKET}/${key}`);
+
+    const signedGetUrlObject = await presigner.presign(
```
agoldis, line 40: @p-8-z This one is going to be stored in the DB and used for all subsequent reads. does it have an expiration?

packages/director/src/screenshots/s3/s3.ts
```diff
@@ -1,78 +1,64 @@
...
+import { fromEnv } from "@aws-sdk/credential-providers";
+import {AssetUploadInstruction} from '@sorry-cypress/common';
+import {HttpRequest} from "@aws-sdk/protocol-http";
+import {S3RequestPresigner} from "@aws-sdk/s3-request-presigner";
+import {parseUrl} from "@aws-sdk/url-parser";
+import {formatUrl} from "@aws-sdk/util-format-url";
+import {Hash} from "@aws-sdk/hash-node";
+import {sanitizeS3KeyPrefix} from '../utils/';
 import {
-  S3_ACL,
```
agoldis, line 5: @p-8-z can you please elaborate - was it deprecated or smth?

packages/director/src/screenshots/s3/s3.ts
```diff
@@ -1,78 +1,64 @@
...
-  S3_BUCKET,
-  S3_FORCE_PATH_STYLE,
-  S3_IMAGE_KEY_PREFIX,
-  S3_READ_URL_PREFIX,
```
agoldis, line 9: @p-8-z I believe some users wants the read url to be completely different from the write S3 bucket because, probably we'd keep it?

agoldis:
@p-8-z thanks a lot for this contribution and the documentation. Please see the inline comments with a few questions,

p-8-z:
@agoldis my first intention was to align url singing so even the Download URL could be sign so the resources are not just publicly available. 
This would allow totally private S3 storage and more privacy.
But yes your concern is right, the signing of Download URL will expire and with signing v4 algorithm the maximum time of expiration is only 7 days. 

So what this means is that with current architecture of storing Download URL after Uploading this is not possible.

My question is, is it possible to split this logic of uploading and downloading?
Director will just upload resource and stores the key into the database then API service would get the key from db and calculate and sign the url on request?

agoldis:
@p-8-z 
Yes - it is possible, you need to generate a new signed URL every time an artifact is returned from the APO

```mermaid
sequenceDiagram
    box Purple Write
    participant cypress
    participant director
    participant DB
    end

    box Green Read
    participant S3
    participant browser
    participant API
    end

    cypress->>director: post artifacts
    director->>DB: save S3 object URL
    director-->>cypress: upload url
    cypress->>S3: upload
    
    browser->>API: get entity with artifacts
    API->>API: generate signed URL
    API-->>browser: signed URL
    browser->>S3: get object via signed URL
    S3-->>browser: object
   
```

PR #858, opened 2023-09-26
fix: make docker-compose.minio.yml work without requiring changes

DamienCassou:
-------

## References

- [ ] I have updated the [documentation](https://github.com/sorry-cypress/gitbook): This change doesn't require any change in documentation. In fact, the current documentation doesn't say anything about changing the docker-compose file so this PR improves the first-time experience by not requiring any change to this file.
- [ ] I have added included automated tests or evidence that it's working: There doesn't seem to be any automated test suite for the docker file.
- [ ] I acknowledge that I have tested that the change is backwards-compatible: Not sure what "backwards-compatible" means in this context.
- [ ] Original issue / feature request / discussion: `<here>`: None.

## Use case

While following the documentation at https://docs.sorry-cypress.dev/guide/dashboard-and-api, I tried to use the docker-compose file without changing it as no change what discussed in the documentation. Unfortunately, the docker-compose file did require some changes for it to work. I believe the new version of the docker-compose file doesn't need any change before it can be used. This should make the documentation easier to follow.

## Example

Not sure what to put here.

### DIFF
--- a/docker-compose.minio.yml
+++ b/docker-compose.minio.yml
@@ -2,8 +2,6 @@ version: '3.6'
 
 # Replace each <mongo-user> with the desired username
 # Replace each <mongo-pass> with the desired password
-# Replace each <minio-ip-address> with the publicly available ip address or url without http(s)
-# Replace <graphql-ip-address> with the publicly available ip address or url without http(s)
 
 services:
   mongo:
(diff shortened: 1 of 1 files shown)


DamienCassou:
I'm closing this PR for now to clean up my todo list. Feel free to reopen when ready to do something about it.

PR #881, opened 2023-11-16
fix: resolve multiple illegal characters in document keys

liamchilds:

## References

- [x] I have updated the [documentation](https://github.com/sorry-cypress/gitbook). PR link 
- [x] I have added included automated tests or evidence that it's working

## Use case

When config keys that contain multiple illegal characters an invalid name error is returned. This is because the current sanitise function only replaces the first example of an invalid character rather than every instance. This PR updates the replace function to a regex that globally searches.

### Before
Screenshot 2023-11-16 at 12 02 06

### After
Screenshot 2023-11-16 at 12 02 22

### DIFF
--- a/packages/director/src/lib/__tests__/fixtures/issue-881.json
+++ b/packages/director/src/lib/__tests__/fixtures/issue-881.json
@@ -2,7 +2,7 @@
   "$schema$schema": {
     "value": "$schema$schema"
   },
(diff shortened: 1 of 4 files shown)


liamchilds:
Hi @agoldis, would you be able to take a look at the above. Hopefully its a non controversial change!

agoldis (approved):

PR #891, opened 2024-01-08
Feature/cypress 13 quickfix

mirobo:
> PRs that do not follow the template will be automatically closed

## References

- [ ] ~I have updated the [documentation](https://github.com/sorry-cypress/gitbook). PR link `<here>`~
- [ ] ~I have added included automated tests or evidence that it's working~
- [ ] ~I acknowledge that I have tested that the change is backwards-compatible~
- [ ] ~Original issue / feature request / discussion: `<here>`~

## Use case

Make sorry-cypress work with Cypress 13 (as good as possible) by removing params that are no longer available in Cypress 13.

Having 'videoUploadOnPasses' produces the error `Error: Cannot return null for non-nullable field CypressConfig.videoUploadOnPasses.` when clicking on a spec result:

image

It is a breaking change or at least a feature-breaking change for Cypress <= 12.

## Example

n/a

PR #892, opened 2024-01-09
Cypress 13 Compatibility for the Dashboard and Director

Roemer:
## References

- [ ] I have added included automated tests or evidence that it's working
- [X] Original issue / feature request / discussion: [Slack](https://sorry-cypress.slack.com/archives/C01346MRQBY/p1704470836208659)

## Use case

A run in Sorry Cypress done with Cypress >= 13 works but when the details of a spec are opened, the following error is shown in the dashboard: 
image
This prevents seeing details about a spec or a failed test in the spec.

The problem is that `videoUploadOnPasses` was removed in Cypress >= 13 but is mandatory in Sorry Cypress.

## Example

See screenshot above

## Notes

I also updated the dev-container specifications a bit so they work again with the newest VSCode and also added some quality of life improvements.

This PR also probably supersedes https://github.com/sorry-cypress/sorry-cypress/pull/891


.gitignore
```diff
@@ -58,9 +57,6 @@ typings/
 # Yarn Integrity file
 .yarn-integrity
 
-# dotenv environment variables file
```
agoldis, line 61: Please restore those

.devcontainer/devcontainer.json
```diff
@@ -4,7 +4,14 @@
   "workspaceFolder": "/workspace",
   "service": "app",
   "forwardPorts": [1234, 8080, 4000, 9000],
-  "settings": {},
-  "extensions": [],
+  "customizations": {
```
agoldis, line 7: Is it necessary?

.gitignore
Roemer: OkI can just include the one from the devcontainer

.devcontainer/devcontainer.json
Roemer: This is the newer format and also makes sure that autoformatting is working correctly in the dev container without modifications or user settings.

Roemer:
I guess I need to re-add the generated files. The tests in the ci don't generate them and fail otherwise. Because the generated files are in gitignore, the tests run with old versions of them which is not optimal.

Roemer:
> I guess I need to re-add the generated files. The tests in the ci don't generate them and fail otherwise. Because the generated files are in gitignore, the tests run with old versions of them which is not optimal.

I actually just added a "generate" script which generates the files and use that before running the tests. This way, the tests always run with the freshly generated files and the files can be removed from the repo.

Roemer:
I think it is ready now, I got everything running in the devcontainer (including minio) and could test it. Feel free to ask if you have any questions.

packages/director/src/app.ts
```diff
@@ -18,14 +18,14 @@ router
   .use(
     PROBE_LOGGER === 'true'
       ? expressPino({
-        logger: getLogger(),
-      })
+          logger: getLogger(),
```
Roemer, line 21: This is only a change because the settings from prettierrc.js were applied correctly.

packages/director/src/app.ts
```diff
@@ -101,12 +115,18 @@ router.post(
   catchRequestHandlerErrors(async (req, res) => {
     const executionDriver = await getExecutionDriver();
     if (executionDriver.id !== 'in-memory')
-      return res.status(405).send('This is only available for in-memory db. Please use the dashboard to set your hooks.');
+      return res
```
Roemer, line 118: This is only a change because the settings from prettierrc.js were applied correctly.

packages/director/src/app.ts
```diff
@@ -101,12 +115,18 @@ router.post(
...
+        .status(405)
+        .send(
+          'This is only available for in-memory db. Please use the dashboard to set your hooks.'
+        );
 
     const { projectId, hooks } = req.body;
     executionDriver.setHooks && executionDriver.setHooks(projectId, hooks);
     getLogger().log(`[hooks] Hooks set for project ${req.body.projectId}`);
-    return res.status(200).send(`Hooks set for project "${req.body.projectId}".`);
+    return res
```
Roemer, line 127: This is only a change because the settings from prettierrc.js were applied correctly.

packages/director/src/app.ts
```diff
@@ -82,6 +82,20 @@ router.put('/instances/:instanceId/stdout', (req, res) => {
   return res.sendStatus(200);
 });
 
+/**
+ * Cypress 13.0.0+ added a delivery notification for artifacts.
+ * The corresponding PR is: https://github.com/cypress-io/cypress/pull/26421
+ * Currently, we just response with an 'OK'
+ */
+router.put('/instances/:instanceId/artifacts', (req, res) => {
```
Roemer, line 90: This is a new endpoint introduced with Cypress 13. It seems to just send an "ok" to the cypress cloud that the screendhots/videos are successfully uploaded now. I think we can just answer with an "ok" for now.

agoldis (approved):

Roemer:
@all-contributors
please add @Roemer for code and infrastucture.

PR #940, opened 2025-02-12
feat: Add optionnal operator to runs query to allow search on createdAt field

ianbabysre:

## References

- [x] I have added included automated tests or evidence that it's working
- [ ] Original issue / feature request / discussion: `<here>`

## Use case
When searching for runs, we currently can't search for runs based on their creation date. This create an issue when calling the API because it generates a huge load and can OOMKill a pod for example.

This feature allows to search for run that were created before or after a certain date passer in the filter, as seen in the screenshots
An operator is mandatory when using createdAt in the filter

## Example

> Submit screenshots / outputs / tests together with the PR.
You can see the runs fetched after a certain date
image
Then, before the same date
image

This screenshot show backward-compatible as the filter is optionnal
image

### DIFF
--- a/packages/api/src/lib/query.ts
+++ b/packages/api/src/lib/query.ts
@@ -2,7 +2,7 @@ export interface AggregationFilter {
   key: string;
   value: string | number | Date;
   operator?: 'gte' | 'gt' | 'lte' | 'lt'; // Make sure operator is included as an optional field
-  like?: string; // If your query includes 'like', this might be a separate field
+  like?: string;
 }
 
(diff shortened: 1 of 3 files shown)


ianbabysre:
Hi @agoldis, hope you're going well

I would like to a feature to SorryCypress. As am not a developper per say, let me know what you think about this implementation.
Especially the modified of the file `packages/api/src/resolvers/index.ts`, it might not be very clean.

Also i have not made a PR to update the documentation as i believe the graphql schema is "enough", let me know if that is necessary !

I'm happy to discuss this with you

